PR #15, opened 2015-02-04
Do not install Genshi by default in the Homebrew installation script

sbesson:
With 5.0.7, Genshi should be bundled as part of the Bio-Formats source.
The `pip install Genshi` command is moved to formulas (4.4.x) requiring
it and will be removed from 5.1.0 eventually.

--depends-on https://github.com/ome/homebrew-alt/pull/71

### DIFF
--- a/homebrew/install_omero51
+++ b/homebrew/install_omero51
@@ -14,6 +14,9 @@ export HTTPPORT=${HTTPPORT:-8080}
 
 cd /usr/local
 
+# Install Genshi (OMERO and Bio-Formats requirement)
+bin/pip install -U genshi
+
 ###################################################################
 # Bio-Formats installation
 ###################################################################
@@ -21,7 +24,6 @@ cd /usr/local
 # Install Bio-Formats
 bin/brew install bioformats51
 VERBOSE=1 bin/brew test bioformats51
-
 
 ###################################################################
 # OMERO installation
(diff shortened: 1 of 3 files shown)


sbesson:
See https://ci.openmicroscopy.org/job/OME-5.0-merge-homebrew/698/

manics:
OME-5.0-merge-homebrew and OME-5.1-merge-homebrew looksfine, do you want to wait for the 4.4 job too?

sbesson:
@manics: if the 5.1 job is fine, I would expect the 4.4 to behave the same (since these 2 scripts are maintaining the Genshi installation). I will monitor the 4.4 job output over the weekend.

homebrew/install_omero51
```diff
@@ -14,6 +14,9 @@ export HTTPPORT=${HTTPPORT:-8080}
 
 cd /usr/local
 
+# Install Genshi (OMERO and Bio-Formats requirement)
+bin/pip install -U genshi
```
rleigh-dundee: Isn't the requirement removed for 5.1 as well now?

sbesson: Absolutely but the 5.1 formula is testing 5.1.0-m3 for now. I will open a PR to remove this when we release 5.1.0-m4

rleigh-dundee:
Overall looks fine and good to merge.  Just one question about if it's still needed for 5.1 or not.

joshmoore:
:+1:

PR #16, opened 2015-02-04
Update OMERO.server to 5.0.8

manics:
Update to 5.0.8 release
Update CentOS instructions to use PostgreSQL 9.4

Will need copying to openmicroscopy/ome-documentation

### DIFF
--- a/linux/setup_omero_ice35.sh
+++ b/linux/setup_omero_ice35.sh
@@ -4,7 +4,7 @@ set -e -u -x
 
 source settings.env
 
-SERVER=http://downloads.openmicroscopy.org/omero/5.0.6/artifacts/OMERO.server-5.0.6-ice35-b53.zip
+SERVER=http://downloads.openmicroscopy.org/omero/5.0.7/artifacts/OMERO.server-5.0.7-ice35-b58.zip
 
 wget $SERVER
 unzip -q `basename $SERVER`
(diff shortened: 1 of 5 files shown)


linux/setup_omero_ice35.sh
```diff
@@ -4,7 +4,7 @@ set -e -u -x
 
 source settings.env
 
-SERVER=http://downloads.openmicroscopy.org/omero/5.0.6/artifacts/OMERO.server-5.0.6-ice35-b53.zip
+SERVER=http://downloads.openmicroscopy.org/omero/5.0.8/artifacts/OMERO.server-5.0.8-ice35-b60.zip
```
sbesson: Use the new `latest/omero/server-ice35.zip` URL ?  :)

rleigh-dundee:
The changes all look reasonable to me.

linux/setup_omero_ice35.sh
manics: How about in the next PR? I'll add separate 50 and 51 scripts.

sbesson: Makes sense. Do you want `latest/omero50/server-ice35.zip` redirects ?

manics: Good question, this is getting towards having stable vs develop in the scripts. Advantage of plain `omero` is the docs will always recommend installing our most recent stable version, OTOH `omero50` is more explicit about what's happening.

PR #17, opened 2015-02-05
Add sleep command to handle race conditions

sbesson:
See https://ci.openmicroscopy.org/job/OME-5.1-merge-homebrew/ history. Many recent instances of the job have been failing while testing the Web client with a `502 bad gateway` error.
This can be reproduced locally by calling

```
omero web restart && 
wget --keep-session-cookies --save-cookies cookies.txt http://localhost:$HTTPPORT/webclient/login/ -O csrf_index.html
```

To prevent race conditions, this PR adds a `sleep` command between the start of the Web client and the test is launched.

### DIFF
--- a/homebrew/install_omero50
+++ b/homebrew/install_omero50
@@ -105,6 +105,7 @@ bin/omero web start
 
 # Test simple Web connection
 brew install wget
+sleep 30
 wget --keep-session-cookies --save-cookies cookies.txt http://localhost:$HTTPPORT/webclient/login/ -O csrf_index.html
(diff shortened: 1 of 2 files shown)


manics:
30 seconds seems quite long.

sbesson:
10 ?

manics:
That sounds reasonable

sbesson:
Pushed - I would propose to wait for a couple of runs of `OME-5.0-merge-homebrew` and `OME-5.1-merge-homebrew` and check that failures due to race conditions disappear.

sbesson:
https://ci.openmicroscopy.org/job/OME-5.0-merge-homebrew/ and https://ci.openmicroscopy.org/job/OME-5.1-merge-homebrew/ have been behaving for the last 10 days with this PR included. Shall we merge?

manics:
:+1:

PR #18, opened 2015-02-17
Linux omero versions

manics:
- Separate 5.0 and 5.1 OMERO server setup scripts (this will allow the docs-autogen job to work on multiple branches without needing multiple branches in omero-install)
- Use `/latest/omero5.X`
- Combine all `install_*.sh` files into one with cmd-line parameters (this is mostly for testing so users shouldn't see the added complexity)

Note Dockerfile testing is only done with 5.0 at present (manually modify the Dockerfiles if you want to test 5.1)


mtbc:
Does `linux/README.md` need updating?

mtbc:
How best to test this? On a fresh image, run the `install.sh`, then some subset (which ?) of the `setup_*.sh` scripts, then with vbox instead of docker is there anything useful in `linux/test` to run or do I then do some subset (?) of the server install walkthrough and see if I have a working server? Is there documentation I'm missing about how to work with these files?

manics:
Good point about the README, updated. Theses scripts are intended to feed be copied into http://www.openmicroscopy.org/site/support/omero5/sysadmins/unix/server-linux-walkthrough.html (though note setup_omero*.sh have been renamed to allow separate 5.0 and 5.1 scripts so that would be one way to test this. `install.sh` is provided as a shortcut for running all the scripts individually.

mtbc:
Ah, so `install.sh` should do everything, okay.

manics:
Yes, though strictly speaking you should check it matches the docs (except the docs are now out of sync...)

mtbc:
What trusted sources are there for the relevant VirtualBox OVAs? I'm guessing it would be imprudent to trust torrents posted at http://virtualboxes.org/. In `ome/team/Software/packages` I can find only a Windows image.

mtbc:
I should check that it doesn't match the docs. :smiley:

mtbc:
(If need be I guess it probably works to download official installer CD/DVD ISOs and run those through VirtualBox, though that's better perhaps done by somebody who has any recent memory of installing CentOS and/or Ubuntu!)

manics:
If it's running in VirtualBox then it should be pretty well sandboxed. Else a plain default install of either OS should work (that's the use case for these scripts). Or team/simon/centos6-selinux/.

mtbc:
Seems to be working pretty well; as far as this PR goes, I can't tell why it's not already **good to merge**.

Some RFEs that are probably outwith this PR (but I'll let you judge, @manics):

* Scripts assume that things like the `omero` user don't already exist, so it might be good someday to adjust so that re-runs can handle the system not being entirely fresh.

* A CentOS apache install has happy OMERO diagnostics but returns a 403 Forbidden; perhaps we need a `setup_apache_centos6_selinux.sh` script?

manics:
Thanks for the review.

> Scripts assume that things like the omero user don't already exist, so it might be good someday to adjust so that re-runs can better handle the system not being entirely fresh.

The idea behind the scripts was to provide a readable but minimal set of instructions to help people setup omero on a new system (i.e. "read through the docs, if they're not clear then hopefully the set of commands in these scripts will help"), so it's explicitly designed not to take any other conditions into account to minimise complexity.

> A CentOS apache install has happy OMERO diagnostics (including that web is running), can do CLI imports, etc., but returns a 403 Forbidden; perhaps we need a setup_apache_centos6_selinux.sh script?

OK, Ill add it to the todo list.

PR #20, opened 2015-02-19
Update windows deps

manics:
Update windows install instructions:

- Ice-3.5.1-b3
- Java 1.8.0_45
- Postgres 9.4.1-3
- OMERO 5.1.1 (using latest redirects isn't straightforward because the unzipped directory name is unknown)
- Use Python wheels for dependencies
- Optional MPlayer
- Remove Python 2.6 and Ice 3.4 scripts (too much effort to maintain both sets of dependencies)

Dependencies can be copied internally from squig:ome/team/simon/omero-install/windows


windows/README.md
```diff
@@ -62,11 +60,25 @@ Finally install and configure OMERO in the usual way:
 
     setup_omero.bat
 
-- omero\OMERO.server-5.0.8-ice35-b60.zip
+- omero\OMERO.server-5.1.1-ice35-b43.zip
+
+
+Optional Prerequisites
```
mtbc: title below has only first words initial-cap

dominikl:
Works fine. But on this occassion, could you add `set PGCLIENTENCODING=UTF8` before calling `psql -hlocalhost -Uomero -fdb.sql omero` in `setup_omero.bat`? Because otherwise you're running into this problem https://www.openmicroscopy.org/qa2/qa/feedback/11012/

manics:
Updated

dominikl:
Looks good. Also already tested it with the line about `PGCLIENTENCODING` included, installation ran through without problems. Good to merge.

PR #21, opened 2015-03-17
Update linux install

manics:
* Replace `$OMERO_HOME` with `$OMERO_SERVER` (shouldn't matter but it's our recommended practice)
* Remove deprecated `omero web config` flags
* Add `omero web clearsessions` to `cron.daily` (and update Dockerfiles to get cron working)
* General cleanups

This will need to be copied to the ome-documentation Linux walkthrough.


linux/omero-init.d
```diff
@@ -23,12 +23,12 @@ prog="omero"
 # Read configuration variable file if it is present
 [ -r /etc/default/$prog ] && . /etc/default/$prog
 
-OMERO_HOME=${OMERO_HOME:-"/home/omero/OMERO.server"}
+OMERO_SERVER=${OMERO_SERVER:-"/home/omero/OMERO.server"}
```
rleigh-dundee: Maybe also add

```
OMERO="$OMERO_SERVER/bin/omero"
```

here, so you can use "$OMERO" in all the commands below?

rleigh-dundee:
Looks fine, though I can't actually test it locally

PR #24, opened 2015-03-24
CentOS6 selinux needs to be configured for Apache as well as Nginx

manics:
IIRC this was suggested by @mtbc a while back.


rleigh-dundee:
This looks fine, please merge.

linux/install.sh
```diff
@@ -49,11 +49,11 @@ su - omero -c "bash -eux setup_$OMEROVER.sh"
 if [ $DISTRO = centos6 ]; then
```
rleigh-dundee: Shell safety note: never use unquoted variable expansion in any place you don't want it to be split into separate tokens

PR #27, opened 2015-04-07
Make sure pip is upgraded

sbesson:
See https://github.com/pypa/pip/commit/368294be59445592cc50551c9a658248912ec10b

Previous version of pip bundled with the Python bottle prevented usage of scc. This commit ensure pip self-upgrades after installation while the bottle is not updated in the upstream repository

### DIFF
--- a/homebrew/install_homebrew
+++ b/homebrew/install_homebrew
@@ -40,6 +40,7 @@ bin/brew doctor
 # Alternately, the system Python can be used but installing Python
 # dependencies may require sudo
 bin/brew install python
+bin/pip install --upgrade pip setuptools
 
 # Tap ome-alt library
 bin/brew tap ome/alt || echo "Already tapped"


sbesson:
See https://ci.openmicroscopy.org/view/Failing/job/OME-5.1-merge-homebrew/91/

manics:
It's obviously fixed the job. Is the plan to revert this when an updated python is released?

sbesson:
@manics: good question. Interestingly, the Homebrew Python formula suggest upgrading pip post installation in its caveats https://github.com/Homebrew/homebrew/blob/master/Library/Formula/python.rb#L287.

I am also fine with keeping this PR open until a new bottle is produced and closing it.

manics:
OK, let's see how long it takes for a new bottle to appear.

manics:
https://github.com/Homebrew/homebrew/commits/master/Library/Formula/python.rb
Looks like the bottle should have pip 6.1.1 now.

sbesson:
Thanks for watching this @manics. Closing this PR

PR #33, opened 2015-07-09
Use exec bash at end of docker test run.sh scripts

manics:
As reported by @joshmoore the `centos6-apache` container was exiting immediately after startup. I'm not sure why but I suspect the parent shell was receiving a termination signal and therefore killing its child bash. Use `exec bash` instead (though if something else is exiting this will probably lead to zombies).

Obviously this is all bad practice for Docker, but these are purely for testing the ome-install scripts.

### DIFF
--- a/linux/test/centos6-apache/run.sh
+++ b/linux/test/centos6-apache/run.sh
@@ -4,4 +4,4 @@ service crond start
 service omero start
 service httpd start
 service omero-web start
-bash
+exec bash
--- a/linux/test/centos6/run.sh
+++ b/linux/test/centos6/run.sh
@@ -4,4 +4,4 @@ service crond start
 service omero start
 service nginx start
 service omero-web start
-bash
+exec bash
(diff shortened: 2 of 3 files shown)


joshmoore:
Testing locally now.

joshmoore:
same behavior?
```
Successfully built 23f5c1ff54fc
Test this image by running docker run -it [...] omero_install_test_centos6-apache
(reverse-i-search)`run': git clone -b fix_docker_run_scripts https://github.com/manics/omero-install
# jamoore@joshuas-mbp:/tmp/omero-install/linux/test$ docker run -it --rm omero_install_test_centos6-apache
Starting postgresql-9.4 service:                           [  OK  ]
Starting crond:                                            [  OK  ]
Starting omero:
# jamoore@joshuas-mbp:/tmp/omero-install/linux/test$ docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
# jamoore@joshuas-mbp:/tmp/omero-install/linux/test$
```

manics:
I've no idea then

PR #35, opened 2015-07-20
make it more obvious that settings.env is not a secvuln

mtbc:
With luck this will stop people finding http://www.openmicroscopy.org/site/support/omero5.1/_downloads/settings.env and reporting its availability as a security vulnerability.


sbesson:
Merging to fix https://ci.openmicroscopy.org/job/OMERO-5.1-merge-docs/ (broken because the autogenerated branch was reverting https://github.com/mtbc/ome-documentation/commit/4cb0b320498978788687cce22919257fda985b81)

PR #37, opened 2015-09-23
Remove 5.0, Switch to WSGI for web, add django

manics:
Switch all web servers to WSGI.
Also added an ~~omero51-merge~~omerodev script (which comes in useful for testing pre-release changes).

Ideally test this after http://downloads.openmicroscopy.org/latest/omero5.1/server-ice35.zip is updated to 5.1.4 (though in theory it should also work with 5.1.3 since WSGI support is included).

Updated: I've also changed the default install to use the `latest/omero5` redirect instead of `5.x`.

PR #41, opened 2015-11-02
Fix nginx Web config target for OMERO 5.2.x

sbesson:
Spotted while reviewing the auto-generation PR, see also https://github.com/ome/omero-install/pull/37#issuecomment-151295246

### DIFF
--- a/linux/setup_omero.sh
+++ b/linux/setup_omero.sh
@@ -20,4 +20,4 @@ psql -h localhost -U "$OMERO_DB_USER" "$OMERO_DB_NAME" < OMERO.server/db.sql
 
 # This is the default in 5.2 so could be left unset
 OMERO.server/bin/omero config set omero.web.application_server wsgi-tcp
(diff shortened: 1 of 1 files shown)


joshmoore:
Testing gh-37, I just got:

```
omero-install/linux/test$ ./docker-build.sh centos6-apache
...
+ OMERO.server/bin/omero config set omero.web.application_server wsgi
+ OMERO.server/bin/omero web config apache --http 80
ERROR: configuration mismatch. omero.web.application_server=wsgi cannot be used with 'omero web config apache'.
The command '/bin/sh -c cd omero-install-test && bash install.sh centos6 apache' returned a non-zero code: 168
```

Related? cc: @manics

sbesson:
Which server is that using? This is the same logic than the one used for the autogeneration scripts in https://github.com/openmicroscopy/ome-documentation/blob/develop/omero/autogen_docs#L36

manics:
@joshmoore probably beacuse setup_omero./sh is still using the latest 5.1 release? Try jusing omerodev.sh instead.

manics:
Thanks @sbesson

PR #49, opened 2015-11-13
Clean Docker shutdown

simleo:
Fixes the problem reported in https://trello.com/c/bJnOwspB/72-omero-install-startup-failure. Apparently, failure to stop `postgresql` before the Docker container is destroyed leaves the DB in an inconsistent state.


joshmoore:
Merging all PRs so someone can open a cleaning/renaming PR.

PR #51, opened 2015-11-14
ubuntu 1504

jburel:
Follow installation of ubuntu1404 (nginx)
install java 1.8 instead of java 1.7. The rest is unchanged.
Apache support will be added following work by @ximenesuk

### DIFF
--- a/linux/dependencies-ubuntu1504.sh
+++ b/linux/dependencies-ubuntu1504.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+apt-get update
+apt-get -y install \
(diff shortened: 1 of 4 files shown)


ximenesuk:
The end of life for 15.04 is January 2016, do we we really need to support this version?

jburel:
This is mainly for testing in case someone asks

ximenesuk:
Yes but then by that argument we must support 12.04 and 15.10. Surely we should only support LTS as we do with other packages?

jburel:
Will check 12.04 and close this one

Issue #46, opened 2015-11-11
Number scripts


joshmoore:
In order to have a sysadmin make use of the scripts, having them number may be beneficial:

 * `01_setup.sh`
 * `02_postgres.sh`
 * etc.

cc: @sbesson @jburel @pwalczysko

manics:
Is #45 https://github.com/manics/omero-install/commit/224d3a98ed8c3a789bfdad9b159cb28b2fca5d5e good enough?

joshmoore:
Certainly better. But replacing the `setup_` names with a counting digit
might significantly increase readability.

ximenesuk:
I don't fully follow your naming suggestion @joshmoore. At the moment we have `setup_blah`. In your example you seem to replacing the `setup` bit for some scripts and the `blah` bit for others?

joshmoore:
@ximenesuk : guess I mostly mean, for any set of scripts, it would be good if a glob would pick things in the right order, something like:

```
dependencies-centos6.sh              --->  01_centos6_deps.sh
dependencies-centos7.sh              --->  01_centos7_deps.sh
dependencies-ubuntu1404.sh           --->  01_ubuntu1404_deps.sh
install-centos7.sh                   --->  install_centos7.sh (or just centos7.sh)
install.sh                           --->  centos6.sh
setup_apache_centos6.sh              --->  05_centos6_apache.sh
setup_centos6_selinux.sh             --->  ???
setup_cron.sh                        --->  ???
setup_nginx_centos6.sh               --->  05_centos6_nginx.sh
setup_nginx_centos7.sh               --->  05_centos7_nginx.sh
setup_nginx_ubuntu1404.sh            --->  05_ubuntu1404_nginx.sh
setup_omero.sh                       --->  04_all_omero.sh
setup_omero_apache.sh                --->  05_centos6_apache.sh
setup_omero_daemon_centos6.sh        --->  06_centos6_daemon.sh
setup_omero_daemon_centos7.sh        --->  06_centos7_daemon.sh
setup_omero_daemon_noweb_centos6.sh  --->  06_centos6_daemon_no_web.sh
setup_omero_daemon_ubuntu1404.sh     --->  06_ubuntu1404_daemon.sh
setup_omerodev.sh                    --->  04_all_omerodev.sh
setup_permissions.sh                 --->  ???
setup_postgres.sh                    --->  03_all_postgres.sh
system_setup.sh                      --->  02_all_setup.sh

```

[...]

PR #53, opened 2015-11-16
Apache centos7

jburel:
Add option to use apache on centos7 
Follow the steps described in https://github.com/ome/omero-install/blob/master/linux/test/README.md
section: Centos 7 testing

To test this PR,
run `bash install-centos7-apache.sh omerodev`

Some files might be renamed depending on the outcome of the current name discussion see https://github.com/ome/omero-install/issues/46

### DIFF
--- a/linux/install-centos7-apache.sh
+++ b/linux/install-centos7-apache.sh
@@ -6,7 +6,7 @@ OMEROVER=omero
 
 for arg in "$@"; do
 	case "$arg" in
-	omero|omerodev|omerodevmerge)
+	omero|omerodev)
 		OMEROVER="$arg"
 		;;
 	*)
--- a/linux/install-centos7-apache.sh
+++ b/linux/install-centos7-apache.sh
@@ -31,7 +31,7 @@ if [ $OMEROVER = omerodev ]; then
 	su - omero -c "bash -eux setup_$OMEROVER.sh"
 fi 
 
-su - omero -c "bash -eux setup_omero_apache.sh"
+su - omero -c "bash -eux setup_omero_apache24.sh"
 bash -eux setup_apache_centos7.sh
 
 #If you don't want to use the systemd scripts you can start OMERO manually:
--- a/linux/install-centos7-apache.sh
+++ b/linux/install-centos7-apache.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+set -e -u -x
+
+OMEROVER=omero
+
+for arg in "$@"; do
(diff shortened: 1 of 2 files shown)


pwalczysko:
Everything seems okay here. The script ran successfully. Started Web and could connect with Insight.

linux/install-centos7-apache.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+set -e -u -x
+
+OMEROVER=omero
```
ximenesuk: I'm a little concerned about this sequence here. While it might make some sense it breaks the pattern used in all the other install scripts.

jburel:
@ximenesuk: that was the point discussed https://github.com/ome/omero-install/issues/46#issuecomment-156997571
we will need the ability to change specify parameter.
we can do that in another PR.
In my view, all install scripts will have to be adjusted after the renaming

linux/install-centos7-apache.sh
```diff
@@ -0,0 +1,44 @@
...
+
+for arg in "$@"; do
+	case "$arg" in
+	omero|omerodev|omerodevmerge)
```
ximenesuk: There is no `omerodevmerge` script corresponding to this option.

jburel: I thought I removed it thanks

ximenesuk:
@jburel no problems with adding options but it would be better discussed and done as a whole after the rename. At the moment the only option it deals with is one that was already dealth with in other scripts and in addition it adds an option that does not work.

jburel:
@ximenesuk: that was discussed this am if I am not mistaken

jburel:
@ximenesuk: removed parameters.
This means that tester will now have to edit the ```install_centos7_apache.sh``` in order to test the PR
if we need to retest it

jburel:
@ximenesuk:
packages installed
```
mod_wsgi-3.4-12.el7_0.x86_64.rpm                                                                                                             |  httpd-2.4.6-31.el7.centos.1.x86_64.rpm
````

ximenesuk:
@jburel thanks for confirming. It's a little confusing as two apache install scripts, one for 2.2 and yours for 2.4, use the same yum installs but maybe what they install is OS-version dependent?

jburel:
Yes the version installed can be different depending on the os version
I have seen the same with ice

ximenesuk:
This is good to merge for me, any further cleaning can be done after the renaming.

PR #54, opened 2015-11-16
Remove Bio-Formats installation from install_omero51 script

sbesson:
This step is now moved to a standalone job and does not need to be executed as part of the OMERO installation.


sbesson:
See https://ci.openmicroscopy.org/job/OMERO-5.1-merge-homebrew/287/

joshmoore:
WFM.

PR #55, opened 2015-11-17
Travis: use omerodev for centos6py27-apache

simleo:
An attempt to fix this:

```
Starting httpd: 91mAH00526: Syntax error on line 72 of /opt/rh/httpd24/root/etc/httpd/conf.d/omero-web.conf:
Either all Options must start with + or -, or no Option may.
```


jburel:
Thanks @simleo

PR #56, opened 2015-11-19
Rename docker install and setup scripts

ximenesuk:
This PR renames the scripts used to build docker images. See https://github.com/ome/omero-install/issues/46 The one slight oddity remaining is that of `centos7` where there is just one `test` directory but two install scripts.

To test each variant should be built and run and a check made that the web server is available.


ximenesuk:
This final commit changes the installation level `README.md`which was out of date due to the splitting of the scripts. It doesn't yet address Centos 7, I'll defer that to someone else.

jburel:
closing this PR so we can create the various branches.

jburel:
README:
should be `bash install_NAMEOFSCRIPT`

jburel:
Adding an example of command on how to run will be useful. e.g. how to be able to access web/install

`docker run --rm -it -p 8080:80 -p 4063:4063 -p 4064:4064 omero_install_test_centos6_nginx`

jburel:
centos6 both apache/nginx okay

jburel:
General comment in dependencies files: we should run a clean post install e.g. `yum clean all`
Probably part of the scripts review. will add a card

manics:
There's no reason to run `yum clean all` on a proper system

jburel:
@manics: ok thanks but we can still clean things that are downloaded

jburel:
ubuntu1404 nginx/apache both okay

jburel:
Debian 8 failure: 

```
E: Failed to fetch http://httpredir.debian.org/debian/pool/main/j/javascript-common/javascript-common_11_all.deb  Error reading from server. Remote end closed connection [IP: 5.153.231.35 80
```

cc @dominikl

jburel:
The following has been removed from the centos7-apache so it cannot be tested using the dev branch (necessary for apache testing)

```
if [ $OMEROVER = omerodev ]; then
    yum -y install python-virtualenv
fi
```

or `python-virtualenv` gets installed regardless of the branch used (with a comment!) in dependencies

jburel:
centos7 both apache and nginx work, Need adjustment to support `omerodev` branch cf. https://github.com/ome/omero-install/pull/56#issuecomment-158415952

jburel:
looking at the ubuntu dependencies file, `python-virtualenv` is installed so we should follow the same strategy for centos7 so the oremodev branch can be used.

ximenesuk:
Sorry, the virtualenv install was killed accidentally with the intention of moving it somewhere else. I'll add to the dependencies and add an example to the README.

jburel:
The debian issue has been introduced in this PR.
I do not have issue when I run `./docker-build.sh debian8` using develop

manics:
You could drop the virtualenv completely and `pip install omego` globally in `*_omerodev.sh`

ximenesuk:
@jburel that's odd as there has been no material change to the debian build in this PR, it still simply uses the ubuntu one.

ximenesuk:
@manics you mean just drop the lines `virtualenv omego` in `step04_all_omerodev.sh` and `step04_centos6_py27_omerodev.sh`?

manics:
@ximenesuk Yes

ximenesuk:
Thanks for the feedback, I'll reconfigure the various files and investigate the debian issue.

hflynn:
This numbered step naming scheme works for me, although it's a bit hard to follow the discussion over 2 different PRs.

jburel:
@ximenesuk:could you re-open a PR so I can start using the new name format in the doc?
We can fix the issues in a follow-up PR

PR #57, opened 2015-11-21
Cleanup Homebrew scripts post branch split

sbesson:
As we are using a multiple branches strategy in sync with openmicroscopy and
ome-documentation, there is no reason to maintain legacy installation scripts.


sbesson:
@ximenesuk @jburel @manics: should we adopt the same naming convention in Homebrew as in linux, e.g. `step00_cleanup_homebrew`, `step_01_install_homebrew`... Also should we go for versionless names i.e. `install_omero`?

jburel:
@sbesson: I will vote for unification and that could make things easier for docs
cc @hflynn

sbesson:
See https://ci.openmicroscopy.org/job/OMERO-DEV-merge-homebrew/33/. For the renaming, I'll likely wait until we have come to an agreement on Linux.

joshmoore:
Perhaps we could also add to the renaming effort:
- versions in script names or not (we currently have one with `51`)
- update to README (which currently skips bioformats)

PR #59, opened 2015-11-24
Rename docker install and setup scripts (on develop)

ximenesuk:
This PR renames the scripts used to build docker images. See #46 and #56 for initial and further discussion. 

The one slight oddity remaining is that of centos7 where there is just one test directory but two install scripts.

To test each variant should be built and run and a check made that the web server is available.


ximenesuk:
The first travis build failed as just one build, `ubuntu1404_nginx`, failed. The second travis build has yet to finish but that build has passed so I think things are fine with travis even if it complains this time.

Edit: but they have passed!

jburel:
Usage of  `omerodev` branch now works and readme fixe
The debian problem will be done in a follow-up PR
cc @dominikl 
Merging

PR #65, opened 2015-12-07
Django1.8

jburel:
Update dependencies to match web requirement.


atarkowska:
Looks good to me

jburel:
Note: When the issue in omero is fixed. Those files will have to be adjusted to use the web requirements files see https://trello.com/c/aFzPitH4/14-usage-of-web-requirements-in-omero-install

PR #67, opened 2015-12-09
Pillow centos7

jburel:
Install pillow using pip
since python-pillow installed version 2.0.0


dominikl:
Good to merge.

joshmoore:
Only minor aside is if that step (also in gh-68) shouldn't possible be refactored out. Merging for now, possible cleanup later.

jburel:
@joshmoore: agreed, will go over a second phase in 5.2.2
since we now have even more steps so we can fully use the include marker in doc

PR #70, opened 2015-12-11
Centos6

jburel:
Make sure that the omerover points to the release version.
When merged, we will need to update the files under walkthrough
cc @hflynn

### DIFF
--- a/linux/docker_virtualenv_ius_apache22.sh
+++ b/linux/docker_virtualenv_ius_apache22.sh
@@ -2,8 +2,7 @@
 
 set -e -u -x
 
-#OMEROVER=omero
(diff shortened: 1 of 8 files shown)


joshmoore:
LGTM. Only thing I can think of to not require switching this on and off would be to use an override:

```
$ echo $HOME
/home/omero
$ echo ${HOME:-foo}
/home/omero
$ echo ${HOMEX:-foo}
foo
```

jburel:
@joshmoore: adding your suggestion to https://trello.com/c/TnPxaDfl/6-install-scripts-parameters

joshmoore:
Merging with travis green.

PR #72, opened 2015-12-13
travis centos6

jburel:
Removing centos6-py27 from travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,9 @@ services:
 env:
   - ENV=centos6_nginx
   - ENV=centos6_apache22
-  - ENV=centos6_py27_nginx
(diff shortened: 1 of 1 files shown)


joshmoore:
Sorry, I missed the discussion. Why are these being removed?

jburel:
Do we still need them to be built since it has been removed from our docs?

joshmoore:
I'd suggest we either build them regularly or delete them.

jburel:
No decision yet. We wait to have everybody back. Probably a discussion for the new year.Closing for now

jburel:
If we keep them around, probably a good idea to provide a ice bundle built against SCL python
I will prepare one when we are done with 5.2.1

PR #79, opened 2016-01-06
Scl ice

jburel:
no longer install ice from source.
Follow the same strategy than the one used for ius python

This should also speed up the travis build

### DIFF
--- a/linux/step01_centos6_py27_deps.sh
+++ b/linux/step01_centos6_py27_deps.sh
@@ -25,9 +25,7 @@ yum -y install \
 	zlib-devel \
 	hdf5-devel \
 	freetype-devel \
-	expat-devel \
-	bzip2-devel \
-	openssl-devel
+	expat-devel
 
 # TODO: this installs a lot of unecessary packages:
 yum -y groupinstall "Development Tools"
--- a/linux/step01_centos6_py27_deps.sh
+++ b/linux/step01_centos6_py27_deps.sh
@@ -36,7 +36,10 @@ set -u
 easy_install pip
 
 export PYTHONWARNINGS="ignore:Unverified HTTPS request"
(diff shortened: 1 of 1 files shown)


dominikl:
Tested the centos6_nginx docker build, omero works fine, but I saw some suspicious output (but I don't think ice was build from source, because the build just took very little time).

```
+ curl -o /etc/yum.repos.d/zeroc-ice-el6.repo http://download.zeroc.com/Ice/3.5/el6/zeroc-ice-el6.repo
[...]

Resolving Dependencies
--> Running transaction check
---> Package ice.noarch 0:3.5.1-1.el6 will be installed
[...]

================================================================================
 Package                Arch     Version                      Repository   Size
================================================================================
Installing:
 ice                    noarch   3.5.1-1.el6                  zeroc-ice    98 k
 ice-python             x86_64   3.5.1-1.el6                  zeroc-ice   445 k
 ice-servers            x86_64   3.5.1-1.el6                  zeroc-ice   3.6 M
 java-1.7.0-openjdk     x86_64   1:1.7.0.91-2.6.2.2.el6_7     updates      26 M
[...]
```

Looks like it's installing ice from the zeroc repo, not from http://downloads.openmicroscopy.org . Is this expected?

jburel:
@dominikl: at which stage did you see that? I do not see any output like the one you mentioned in the console

dominikl:
Just ran it again, see attached output (line 431 and following)

[docker-build.txt](https://github.com/ome/omero-install/files/81330/docker-build.txt)

dominikl:
Used the wrong build previously. Running the docker build of centos6_p27 works fine, ice from ome downloads is used, omero server and omero.web runs without problems. Good to merge.

joshmoore:
:thumbsup:

PR #81, opened 2016-01-08
Readme

jburel:
Update readme

### DIFF
--- a/linux/README.md
+++ b/linux/README.md
@@ -11,11 +11,15 @@ Copy the files from this directory, then run one of the install scripts,
 	bash install_centos6_nginx.sh
 	bash install_centos6_py27_apache24.sh
 	bash install_centos6_py27_nginx.sh
+	bash install_centos6_py27_ius_apache22.sh
+	bash install_centos6_py27_ius_apache24.sh
+	bash install_centos6_py27_ius_nginx.sh
 	bash install_debian8_nginx.sh
+	bash install_debian8_apache24.sh
 	bash install_ubuntu1404_apache24.sh
 	bash install_ubuntu1404_nginx.sh
 
-corresponding to the required OS and web platform. Note that for Centos 6
+corresponding to the required OS and web platform. Note that for CentOS 6
 there are separate scripts for Python 2.7, the default is Python 2.6.
 
 Usernames and passwords can be customized in `settings.env`.
--- a/linux/test/README.md
+++ b/linux/test/README.md
@@ -17,7 +17,7 @@ For example:
 See `docker run --help` for more information on these and other options
 for running docker images.
 
-CentOS 7 cannot be tested in this way as systemd doesn't fully work, see below.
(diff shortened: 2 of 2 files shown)


linux/test/README.md
```diff
@@ -5,19 +5,16 @@ This directory contains Dockerfiles for testing the installation walkthroughs.
...
+    docker run --rm -it -p 8080:80 -p 4063:4063 -p 4064:4064 omero_install_test_centos6_apache22
 
 See `docker run --help` for more information on these and other options
 for running docker images.
 
-Ubuntu 14.04 using apache needs the web configuration to be modified to
-enable the OMERO web server to run, see below.
-
-Centos 7 cannot be tested in this way as systemd doesn't fully work, see below.
+CentOS 7 cannot be tested in this way as systemd doesn't fully work, see below.
```
ximenesuk: Since CentOS 6 ius now has further instructions below can this sentence be expanded to reflect that?

ximenesuk: Also, while you are modifying this file could you change the couple of instances of systemd to `systemd`?

manics: Systemd is the name of a suite of tools, not a command, though Wikipedia seems confused on the correct capitalisation https://en.wikipedia.org/wiki/Systemd

jburel: @ximenesuk I was planning to do in the centos6 ius PR 
the same applies to the new parameter for webapp. 
but if you are happy to review the changes in this PR even if it covers the other two I am fine with that

ximenesuk: @manics yes, I wasn't sure here whether it refers to the daemon or the bundle? Maybe use bold?

@jburel I'm fine to review the wider changes in the PR

jburel:
@ximenesuk: I have more commits to come in the next few days so I removed the PR from standup list

linux/README.md
```diff
@@ -11,11 +11,15 @@ Copy the files from this directory, then run one of the install scripts,
 	bash install_centos6_nginx.sh
 	bash install_centos6_py27_apache24.sh
 	bash install_centos6_py27_nginx.sh
+	bash install_centos6_py27_ius_apache22.sh
+	bash install_centos6_py27_ius_apache24.sh
+	bash install_centos6_py27_ius_nginx.sh
 	bash install_debian8_nginx.sh
```
dominikl, line 17: The centos7 scripts could be listed, too, just to be complete.

dominikl, line 17: Ah, forgot that CentOS 7 is a bit special and cannot not be installed fully automatic, so might be fine to not list it here explicitly.  But maybe worth adding a note about it, e. g. referring to the CentOS 7 remark in `linux/test/README.md`?

jburel, line 17: CentOS 7 does not follow the same pattern, that's why it has not been listed.

dominikl:
Ok, good to merge (my comment above is not a blocker) :+1:

PR #91, opened 2016-02-15
Docker installation

jburel:
Disable interactive when installing docker-engine
This should force the config to be overwritten
cf https://github.com/ome/ome-docker/pull/30


jburel:
Everything is green 
cc @sbesson @joshmoore

PR #93, opened 2016-03-09
Doc generate

jburel:
First step towards improvement of the doc installation
This "autogenerate" script will add the content of the various steps (some sections are excluded e.g. docker specific).
This will generate `walkthrough_OS.sh` files that we can then use in the omero-doc
The `walkthrough_OS.sh` files will have the `recommended` versions install e.g. java 1.8, etc. and should be used as detailed guideline (not as a runnable script!). The files should contain the markers so we can include the code snippets in the pages.
The `walkthrough_OS.sh` files will contain both Nginx and Apache installation steps.
This can easily be modified to use a `webserver` parameter

Markers have been added to the step files so we do not have hard-coded line number in the autogenerate (excluding 2 since all scripts will have `#!/bin/bash` on the first line)

usage: 
- Run  `OS=centos6 bash autogenerate.sh` to generate the `walkthrough_centos6.sh` file
- OS values should be one of the following: 
  - centos7 (default), centos6_py27, centos6_py27_ius, ubuntu1404, debian8, centos6

cc @hflynn @joshmoore @sbesson

### DIFF
--- a/linux/README.md
+++ b/linux/README.md
@@ -24,3 +24,15 @@ there are separate scripts for Python 2.7, the default is Python 2.6.
 
 Usernames and passwords can be customized in `settings.env`.
 
+Documentation generation
+========================
+To generate a walkthrough file corresponding to a given OS, run the
+`autogenerate.sh` script, the OS is specified as a paramater e.g.
+	
+	OS=ubuntu1404 bash autogenerate.sh
+
+The walkthrough file is used for the omero documention e.g.
+http://www.openmicroscopy.org/site/support/omero5.2/sysadmins/unix/server-linux-walkthrough.html
+but it should not be executed.
+
+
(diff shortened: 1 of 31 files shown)


jburel:
Example file (txt used only to upload file): cc @hflynn 
[walkthrough_centos7.txt](https://github.com/ome/omero-install/files/165471/walkthrough_centos7.txt)

pwalczysko:
Created the walkthrough...sh files for
centos7 (default), centos6_py27, centos6_py27_ius, ubuntu1404, debian8, centos6
Opened the files and checked them briefly in text editor. 
They seem okay.
Is any more thorough check necessary ?

jburel:
Discussed with @joshmoore and @sbesson, we might merge it as such
Further adjustment will probably required.
Long term will be to move the sysadmin doc under omero-install
and reference it from the main doc. cc @hflynn

joshmoore:
WFM.

PR #96, opened 2016-04-07
debian 8 nginx

jburel:
Install nginx version 1.8 on debian 8
There is a change in the configuration between 1.6 (cf. ubuntu install) and 1.8

To test:
- Build the image i.e. `./docker-build.sh debian8_nginx/`
- Then run `docker run --rm -it -p 8080:80 -p 4063:4063 -p 4064:4064 omero_install_test_debian8_nginx`
- Check nginx version i.e. `nginx -v`
- Check that the login page of web is up.

cc @aleksandra-tarkowska

### DIFF
--- a/linux/step05_debian8_nginx.sh
+++ b/linux/step05_debian8_nginx.sh
@@ -5,6 +5,11 @@ cp setup_omero_nginx.sh ~omero
 #end-copy
 
 #start-install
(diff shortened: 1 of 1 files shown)


dominikl:
Works fine, nginx 1.8.1 gets installed, and OMERO.web is up and running. :+1:

PR #102, opened 2016-04-14
Ice36 release

jburel:
Install ice 3.6 on all os
git is installed but will be removed when we can use omego.
the `step04_all_omero.sh` file also need to be modified when the openmicroscopy PR is merged.

This is not ready to be merged

### DIFF
--- a/linux/step01_centos7_ice_deps.sh
+++ b/linux/step01_centos7_ice_deps.sh
@@ -29,7 +29,4 @@ elif [ "$ICEVER" = "ice36" ]; then
 	yum -y install ice-all-runtime ice-all-devel
 
 	pip install zeroc-ice
-
-	echo /opt/Ice-3.6.2/lib64 > /etc/ld.so.conf.d/ice-x86_64.conf
-	ldconfig
 fi
--- a/linux/step01_centos7_ice_deps.sh
+++ b/linux/step01_centos7_ice_deps.sh
@@ -15,4 +15,32 @@ elif [ "$ICEVER" = "ice35-devel" ]; then
 	http://download.zeroc.com/Ice/3.5/el7/zeroc-ice-el7.repo
 
 	yum -y install ice ice-python-devel ice-java-devel ice-servers
+elif [ "$ICEVER" = "ice36" ]; then
+	curl -o /etc/yum.repos.d/zeroc-ice-el7.repo \
+	http://download.zeroc.com/Ice/3.6/el7/zeroc-ice-el7.repo
+	yum -y groupinstall "Development tools"
+	yum -y install openssl-devel bzip2-devel expat-devel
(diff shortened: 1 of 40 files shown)


linux/step01_centos7_ice_deps.sh
```diff
@@ -15,4 +15,18 @@ elif [ "$ICEVER" = "ice35-devel" ]; then
 	http://download.zeroc.com/Ice/3.5/el7/zeroc-ice-el7.repo
 
 	yum -y install ice ice-python-devel ice-java-devel ice-servers
+elif [ "$ICEVER" = "ice36" ]; then
+	cd /etc/yum.repos.d
+	wget https://zeroc.com/download/rpm/zeroc-ice-el7.repo
+
+	#for omero, to remove when we do not install from git
+	yum -y install git
```
atarkowska: is there any reason for installing get as a part of ice dependencies? why not do it in https://github.com/ome/omero-install/pull/102/files#diff-49a05e9c183b6b84aa966891a004eb37R30

linux/test/test_services.sh
```diff
@@ -42,14 +42,14 @@ docker exec -it $CNAME /bin/bash -c 'd=10; \
         done'
 
 #check OMERO.server service status
-docker exec -it $CNAME /bin/bash -c "service omero status -l"
+docker exec -it $CNAME /bin/bash -c "systemctl status omero.service -l"
```
atarkowska: if https://github.com/ome/ome-docker/pull/36/files is in, the above can be reverted

manics:
Is there an ice36 CI build? omego should already work with it, though you might need to add a labels argument.

jburel:
@manics: we don't have one yet, as soon as it is working I will revert and use labels.
That's the plan cf. description

linux/step01_centos7_ice_deps.sh
jburel: I put it in ice 36 section b/c of omero installation, It will be removed as soon as I use label as indicated in the description
It is not done in step04_omero since all os scripts now use it.

linux/test/test_services.sh
jburel: Reverted the changes in test waiting for the upstream change

manics:
@jburel Thanks, I misunderstood the description- I thought you meant changes were needed to omego

jburel:
FYI: Debian currently not passing. Looking into it.

jburel:
due to constant failure on debian, building from source for now

jburel:
all, expect centos7_apache, are green cf https://travis-ci.org/ome/omero-install/builds/123329750.
Reverting commit.

jburel:
note that the omero set-up will be adjusted prior to release.
Currently pointing a build on breaking

PR #113, opened 2016-08-15
debian 8

jburel:
Update java8 headless version
This should hopefully fix the build issue.

cc @mtbc


mtbc:
Must we fix the version at all?

atarkowska:
great eventually passing, could we merge that please to test with other PRs?

PR #115, opened 2016-08-23
autogen fix

jburel:
Copy correct file.
The correct file was not copied
to test:
- go to linux directory
- run `bash autogenerate.sh`
- Check that there is no `omero-.env` after
- Also check that the correct commands are displayed to install postgres on debian 8
  `
  #start-step04: As the omero system user, install the OMERO.server
  #start-copy-omeroscript
  `
  cc @mtbc

### DIFF
--- a/linux/autogenerate.sh
+++ b/linux/autogenerate.sh
@@ -218,7 +218,7 @@ echo -en '\n' >> $file
 echo "#start-step04: As the omero system user, install the OMERO.server" >> $file
 if [ $OS = "centos6_py27" ] ; then
 	echo "#start-copy-omeroscript" >> $file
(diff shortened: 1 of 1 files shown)


mtbc:
Only env files are,

```
omero-centos6py27.env
omero-centos6py27ius.env
omero-ice36.env
settings.env
```

so no `omero-.env`.

mtbc:
These changes look good. Does CI test that it still works for Debian?

jburel:
the script is not run on commits base, we usually ran it to upgrade the walkthrough for the doc i.e. https://github.com/openmicroscopy/ome-documentation/blob/develop/omero/autogen_docs

Note that debian8 is not included in the doc that's why it is moved to community-docs repo.

joshmoore:
Restarted due to

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

```

PR #116, opened 2016-08-24
centos6 omero user

jburel:
Make sure all the lines to create the "omero" system users are copied to the walkthrough

to test
- Go to linux directory
- run `bash autogenerate.sh`
- Check that in the `walkthrough_centos6_py27.sh` file, the following line `chmod a+X ~omero` is present


mtbc:
Good to merge.

PR #117, opened 2016-09-08
avoid stderr when generating webserver config

atarkowska:
This PR improve scripts handling web server config to only save stdout


joshmoore:
Was this a breaking change?

atarkowska:
that was a mistake in the code, I originally used `ctx.out` rather then `ctx.err`. I just used wrong standard streams. That is not needed

PR #120, opened 2016-09-23
Homebrew uninstall

sbesson:
Following some the upgrade of the brew node by @rleigh-codelibre, some assumptions of the scripts within the job environment do not hold anymore.

This PR:
- uses the official uninstall script instead of our hard-coded `rm -rf` - this should be subjected to the same permissions as the install script (meaning `hudson` need to sudo)
- stores the generated sql script under `~` to fix permission issues under `/usr/local`
- reinitializes the PSQL database cluster using `initdb` to fix missing directories (see [1](https://coderwall.com/p/rjioeg/fixing-postgres-homebrew-install-after-yosemite-upgrade), [2](https://foobartel.com/articles/yosemite-upgrade-postgresql/), [3](http://apple.stackexchange.com/questions/209654/os-x-seems-to-delete-empty-directories))

### DIFF
--- a/osx/step02_omero.sh
+++ b/osx/step02_omero.sh
@@ -7,7 +7,7 @@ set -x
 
 export PATH=/usr/local/bin:$PATH
 export OMERO_DATA_DIR=${OMERO_DATA_DIR:-~/OMERO.data}
-export PSQL_SCRIPT_NAME=${PSQL_SCRIPT_NAME:-OMERO.sql}
+export PSQL_SCRIPT_NAME=${PSQL_SCRIPT_NAME:-~/OMERO.sql}
 export ROOT_PASSWORD=${ROOT_PASSWORD:-omero_root_password}
 export ICE=${ICE:-3.5}
 export HTTPPORT=${HTTPPORT:-8080}
--- a/osx/step02_omero.sh
+++ b/osx/step02_omero.sh
@@ -7,7 +7,6 @@ set -x
 
 export PATH=/usr/local/bin:$PATH
 export OMERO_DATA_DIR=${OMERO_DATA_DIR:-~/OMERO.data}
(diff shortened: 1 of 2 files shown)


sbesson:
The last commit seems to be sufficient to let https://ci.openmicroscopy.org/job/OMERO-DEV-merge-homebrew/287/ pass. Listing for review.

jburel:
The will require a documentation PR to update the `step02_omero.sh`

sbesson:
Yes, I realize now that the documentation page is still hardcoded and needs to be maintained in sync. Let's make sure we agree on the code modifications first.

osx/step02_omero.sh
```diff
@@ -7,7 +7,7 @@ set -x
 
 export PATH=/usr/local/bin:$PATH
 export OMERO_DATA_DIR=${OMERO_DATA_DIR:-~/OMERO.data}
-export PSQL_SCRIPT_NAME=${PSQL_SCRIPT_NAME:-OMERO.sql}
+export PSQL_SCRIPT_NAME=${PSQL_SCRIPT_NAME:-~/OMERO.sql}
```
joshmoore: Perhaps another option is a temp file though no file is strictly needed:

```
omero db script --password $ROOT_PASSWORD -f - |\
    psql -h localhost -U db_user omero_database
```

sbesson: That was the original idea of use the current directory (under `/usr/local`) as it was getting wiped out by the cleanup step. I can use `/tmp/OMERO.sql` if that makes sense.

joshmoore:
- Job looks good.
- Travis failures are unrelated to this PR. 
- Happy to skip the `-f-` usage.

Assuming there's nothing that needs doing here for docs, merge when ready.

joshmoore:
Happy to merge with next green build.

joshmoore:
Passed yesterday and looking good today.

PR #123, opened 2016-10-26
upgrade django-redis

atarkowska:
When using redis we have to use newer supported package https://trello.com/c/GIJQA0kg/177-redis-out-of-experimental


jburel:
we probably don't need to update the "web" section if we are moving everything related to web to omero-webinstall

atarkowska:
@jburel you are right, but I would prefer to keep repos in sync if that is ok

PR #129, opened 2017-01-08
install pillow from rpm

atarkowska:
following a big effort of fixing https://trello.com/c/jZjPZf8K/216-attributeerror-numpy-ndarray-object-has-no-attribute-tobytes#comment-58206ac0ad940307d3bd6716, installation and doc can be cleaned up

tag as 5.3.0-M?, RC?


joshmoore:
Wow. This looks great. What all needs to be tested?

atarkowska:
list of tasks is in https://github.com/openmicroscopy/devspace/pull/63 <- is an entry point

jburel:
each time we make a milestone we tag omero-install current tag is v5.3.0-m6

atarkowska:
So we should do the same for docker image I think, and maintain them all!

atarkowska:
@jburel so looks like all is good https://10.0.51.143:8443/job/Trigger/1/ blue in a first build 

```
python-pillow.x86_64                      2.0.0-19.gitd1c6db8.el7        @base  
```

linux/step01_centos6_deps.sh
```diff
@@ -3,12 +3,8 @@
 yum -y install \
 	python-pip python-devel python-virtualenv \
 	python-yaml python-jinja2 \
-	numpy scipy python-matplotlib Cython \
+	numpy scipy Cython \
```
aleksandra-tarkowska: does anyone knows what `scipy` and `Cython` are for?

atarkowska:
out of scope of this PR, does anyone knows what `scipy` and `Cython` are for?

jburel:
``linux/requirements.txt`` is probably no longer needed. 
It is currently empty

jburel:
Scipy is a dependency of OMERO.web
cf. ``components/tools/OmeroWeb/omeroweb/webclient/views.py``

atarkowska:
@jburel I left requirements.txt on purpose in case we need to add anything in the future. I would prefer to leave it as it is

jburel:
@aleksandra-tarkowska: I am happy with that

atarkowska:
thx, true https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webclient/views.py#L2925

ok so no more cleanups :)

jburel:
Thanks @aleksandra-tarkowska. Merging

PR #130, opened 2017-01-24
Ice36 default

jburel:
Make ice 3.6 the default value
This also change the output of the "autogen" script
the recommended version is now ice 3.6

To test:
 * travis should be green
 * run ``autogen.sh``
 * Check that that before and after the commands to install ice 3.6 there are 
``#start-recommended-ice`` and ``#end-recommended-ice``


dominikl:
Good to merge 👍

PR #133, opened 2017-01-25
centos7 fix

jburel:
Use the new upstream image
This should fix the failure due to "locale" issues

See also https://github.com/openmicroscopy/omero-ssh-c7-docker/releases/tag/0.1.0-1

Travis should hopefully be green for centos7


jburel:
All green cc @sbesson @joshmoore

PR #137, opened 2017-02-03
Bump pg version

jburel:
This PR bumps the version of postgresql to the latest point release

See 
https://yum.postgresql.org/repopackages.php


rleigh-codelibre:
The updated links are all valid.  Please merge.

PR #138, opened 2017-02-08
add git

jburel:
Install git so OMERO-DEV version can be used

Due to the addition of omero-marshal it is currently not possible to build with ``OMERO-DEV-*``

### DIFF
--- a/linux/step01_centos6_init.sh
+++ b/linux/step01_centos6_init.sh
@@ -3,4 +3,4 @@
 yum -y install epel-release
 
 # installed for convenience
-yum -y install unzip wget bc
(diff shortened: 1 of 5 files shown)


jburel:
Build https://travis-ci.org/ome/omero-install/builds/200318643 tested the installation of OMER0-DEV-latest

sbesson:
Is that a temporary workaround? I would assume for the final release of OMERO 5.3.0, we will consume  a proper Pypi release of `omero-marshal` rather than a Git clone.

jburel:
mainly to be able to build now if someone needs it

PR #139, opened 2017-02-09
Debian java

jburel:
Due to a recent bug see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=851667
The installation of Java on Debian always fails (see recent PRs)
This PR fixes the issue

Check that travis is green


mtbc:
This change certainly makes sense.

manics:
Is the plan to keep this or revert it when the upstream bug is fixed?

jburel:
The plan is to only use ansible playbook. The changes made in this repo at the moment are mainly for the documentation.
This is mainly to have travis green, we can revert when it is fixed upstream

dominikl:
Looks good, ran a quick docker build, ` ca-certificates-java` and `openjdk-8-jdk` were installed successfully. 👍

PR #146, opened 2017-03-16
Latest requirements

jburel:
* Update the version to 5.3
 * Use the latest requirements file to match the rest of the doc


.travis.yml
```diff
@@ -16,7 +16,7 @@ before_install:
   - sudo apt-get -y install shunit2
 script:
     - cd linux/test
-    - ./docker-build.sh $ENV
+    - OMEROVER=OMERO-DEV-latest ./docker-build.sh $ENV
```
joshmoore, line 19: The idea is to drop this once 5.3.0 is released?

linux/step05_centos6_py27_ius_nginx.sh
```diff
@@ -30,11 +30,16 @@ set -u
...
 
-# introduced in 5.2.0
+file=~omero/OMERO.server/share/web/requirements-py27-all.txt
+
+# introduced in 5.3.0
 if [ -f $file ]; then
 	#start-latest
 	/home/omero/omeroenv/bin/pip2.7 install -r $file
 	#end-latest
+elif $(is_less_than $OMEROVER 5.3); then
```
joshmoore: :+1: for trying to keep backwards compatibility but I guess there's a question of whether or not we try to do so. Are we basically saying `master` is now 5.3 and therefore we either branch off `dev_5_2` or tell anyone who wants to to use a tag?

joshmoore (commented):
Few questions, largely clarifications of the "major-version-bump" process for this repo.

cc: @sbesson

.travis.yml
jburel: only to have travis to pass otherwise m9 will be used with contain omero-marshal dependencies installed using git

linux/step05_centos6_py27_ius_nginx.sh
jburel: the various switches in place are mainly  there if someone needs to "build" a 5.1, etc. server
happy to remove them since previously we did not have branches for that repo only develop

jburel: happy to remove the various versions check since we now have dev_5_1 and dev_5_2 branches
this will also simplify the code and remove the need of the version in utils

joshmoore:
re: c93e05afd7e7ba51962e4c97a340526cf74524b8 -- I'm not sure how it is to keep the `5.3`-specific parts while dropping the `5.2`- and `5.1`-specific parts, but having this repo "know" that it works with 5.3 could still be a good thing, so that when `/latest/omero/` moves to 5.4 (or later) this repo will keep working. Then we can explicitly update this repo to support 5.4 or choose not to so that it keeps being 5.3 only.

sbesson:
Although it might mean regular review, I agree with https://github.com/ome/omero-install/pull/146#issuecomment-288028623 especially since the content of this repository is directly exposed in the public documentation for  OMERO 5.3.x.
More generally, I think that as long as in our OMERO versioning policy remains to allow breaking requirements changes as part of a minor version increment, we should stay on the safe side and favor the usage of minor series redirects i.e. `latest/omeroX.Y`.

dominikl:
Tried a docker build of `ubuntu1604_nginx`, got a:
```
+ file=/home/omero/OMERO.server/share/web/requirements-py27-all.txt
+ pip install -r /home/omero/OMERO.server/share/web/requirements-py27-all.txt
Could not open requirements file: [Errno 2] No such file or directory: '/home/omero/OMERO.server/share/web/requirements-py27-all.txt'
```

jburel:
@dominikl: This is b/c you did not specify the omero version i.e. it will have been built with the latest official release 5.2.7

dominikl:
Ah, of course, thanks.

jburel:
@joshmoore @sbesson added 5.3 so it is explicit
same will need to be done in https://github.com/ome/omeroweb-install/pull/12
cc @aleksandra-tarkowska

dominikl:
Tried a couple of docker builds yesterday, worked fine. Will try some again today, with the latest commits.

dominikl:
Tried a `ubuntu1604_nginx` and a`centos6_py27_nginx` docker build, no problems 👍

linux/step05_ubuntu1404_nginx.sh
```diff
@@ -2,14 +2,10 @@
...
-source utils.sh
-
 #start-copy
 cp setup_omero_nginx.sh ~omero
 #end-copy
 
-p=nginx
-
 #start-install
 # require to install more recent version of nginx
```
mtbc: maybe "required"

linux/step05_ubuntu1404_nginx.sh
```diff
@@ -2,14 +2,10 @@
...
 # w/o the version installed is 1.4.6
```
mtbc: maybe expand "w/o" to "without this,"

jburel:
The change in ``travis.yml`` will be reverted before tagging

PR #152, opened 2017-04-04
ice 3.6 

jburel:
Make sure ice 3.6 is the default
When the script is used by itself, it defaults to ice 3.5


dominikl:
👍  Good to merge.

PR #154, opened 2017-06-13
Set the locale

jburel:
en_US.UTF-8 no longer available this leads to an error when starting pg


dominikl:
Tested with docker, no problems, install, omero startup, web, everything alright 👍

PR #157, opened 2017-06-14
Omero version

jburel:
#it is no longer possible to retrieve the version of omero server with ice 3.5 from the download page
The version will be installed using omego
This requires a bump of omego i.e.  version 0.6.0

To test run for example
``ICEVER=ice35 ./docker-build.sh ubuntu1604_nginx``


dominikl:
Seems to work. Downloads omego 0.6.0 and calls it to get OMERO:
```
...
Collecting omego==0.6.0
Downloading omego-0.6.0.tar.gz
...
/home/omero/omeroenv/bin/omego download --ice 3.5 --branch latest server
...
```
👍

jburel:
Thanks for testing

PR #160, opened 2017-08-04
Remove scipy

jburel:
No longer install scipy
The code using scipy has now been removed from omero
See https://github.com/openmicroscopy/openmicroscopy/pull/5412

Check that travis is green


dominikl:
Looks ok. Tried some local builds, works fine with centos, but got the same `fatal error: Python.h: No such file or directory` issues with debian8 and ubuntu14_04 builds like travis.

jburel:
ice error by the look of it.
not related to the change. I have restarted again the job

jburel:
failure not related to the change merging for m3

PR #163, opened 2017-09-05
Update travis

jburel:
Remove
 
* group: deprecated-2017Q2
* use services docker i.e. do not install docker engine

To test:
  * open a travis build and check that 14.04 (trusty) has been used
  * check that travis is green


joshmoore:
LGTM

PR #165, opened 2017-09-13
Ice 3.6.4

jburel:
Change to check the bump to version 3.6.4

Check that ice 3.6.4 is installed in the travis build
check C7 debian9 ubuntu1604

### DIFF
--- a/linux/step01_centos6_py27_ice_deps.sh
+++ b/linux/step01_centos6_py27_ice_deps.sh
@@ -27,9 +27,7 @@ if [[ "$ICEVER" =~ "ice35" ]]; then
     #end-supported
 elif [ "$ICEVER" = "ice36" ]; then
 	#start-recommended
-	cd /etc/yum.repos.d
-	wget https://zeroc.com/download/rpm/zeroc-ice-el6.repo
-	curl -sL https://zeroc.com/download/Ice/3.6/el6/zeroc-ice3.6.repo > \
+    curl -sL https://zeroc.com/download/Ice/3.6/el6/zeroc-ice3.6.repo > \
     /etc/yum.repos.d/zeroc-ice3.6.repo
 
 	yum -y install gcc-c++
--- a/linux/step01_centos6_py27_ice_deps.sh
+++ b/linux/step01_centos6_py27_ice_deps.sh
@@ -29,6 +29,8 @@ elif [ "$ICEVER" = "ice36" ]; then
 	#start-recommended
 	cd /etc/yum.repos.d
 	wget https://zeroc.com/download/rpm/zeroc-ice-el6.repo
+	curl -sL https://zeroc.com/download/Ice/3.6/el6/zeroc-ice3.6.repo > \
+    /etc/yum.repos.d/zeroc-ice3.6.repo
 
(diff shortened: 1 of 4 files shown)


mtbc:
> check C7 debian9 ubuntu1604

... in the Travis logs for the matrix job?

mtbc:
CentOS 7:
```
---> Package ice-all-devel.x86_64 0:3.6.4-1.el7 will be installed
...
---> Package ice-all-runtime.x86_64 0:3.6.4-1.el7 will be installed
...
 ice-all-devel       x86_64    3.6.4-1.el7         zeroc-ice3.6            13 k
 ice-all-runtime     x86_64    3.6.4-1.el7         zeroc-ice3.6            14 k
```

Debian 9:

```
Setting up libzeroc-ice3.6:amd64 (3.6.3-5) ...
...
Setting up zeroc-ice-all-runtime (3.6.3-5) ...
...
Successfully installed zeroc-ice-3.6.4
```

Ubuntu 16.04:

```
Setting up libzeroc-ice3.6:amd64 (3.6.4-1000) ...
...
Setting up zeroc-ice-all-runtime (3.6.4-1000) ...
...
Successfully installed zeroc-ice-3.6.4
```

CentOS 7 looks good. On Debian we seem to install 3.6.3 via the package manager then 3.6.4 with pip. On Ubuntu we seem to install 3.6.4 via the package manager then 3.6.4 again with pip.

jburel:
It might not yet be available on debian since we do not specify a version

commands run
```
apt-get -y install libssl-dev libbz2-dev libmcpp-dev libdb++-dev libdb-dev libdb-java
apt-get -y install zeroc-ice-all-runtime
pip install "zeroc-ice>3.5,<3.7"
```

mtbc:
Maybe we don't need that middle one at all for either Debian or Ubuntu if we're installing via pip anyway?

jburel:
pip only installs the Ice Python extension
so you need the rest

mtbc:
Ah, right. Debian's unstable repository does offer 3.6.4 but I don't know how much we want to get into fetching from there nor how optimistic we want to be about it ever making it to a stable update for stretch.

linux/step01_centos6_py27_ice_deps.sh
```diff
@@ -29,6 +29,8 @@ elif [ "$ICEVER" = "ice36" ]; then
 	#start-recommended
 	cd /etc/yum.repos.d
 	wget https://zeroc.com/download/rpm/zeroc-ice-el6.repo
```
joshmoore: This `wget` can be omitted, no?

jburel: oops 👍 thanks for spotting it

jburel:
For the last 2 commits Check that in the travis build for ``centos6_py27_nginx`` and ``centos6_py27_ius_nginx`` ice 3.6.4 is intalled

mtbc:
Note that curl offers `-o` but current redirection is fine too and https://github.com/ome/omero-install/pull/165#issuecomment-330157697 is as expected. :+1:

mtbc (approved):

PR #166, opened 2017-09-15
Update url

jburel:
Replace http by https
Note that the web apps installation is legacy
i.e. all apps can now be installed using pip. This has not been reflected in this code
The web apps code is not used in the doc and the apps are deployed daily


joshmoore:
:+1:

PR #167, opened 2017-09-15
Rename variables

jburel:
Rename variable to match name in omeroweb-install
To avoid confusion when reading the installation page
see https://github.com/openmicroscopy/openmicroscopy/issues/5508


dominikl:
Looks good, matches documentation now. Good to merge 👍

PR #169, opened 2017-09-20
Debian9 tables

jburel:
tables version 3.3 currently does not work
install directly from pypi This should install version 3.4.x
see https://github.com/openmicroscopy/openmicroscopy/pull/5481

Make sure travis is green

cc @mtbc @joshmoore


linux/step01_debian9_deps.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+apt-get -y install \
+	python-{pip,virtualenv,yaml,jinja2}
+
+pip install --upgrade pip
+
+# python-tables will install tables version 3.3
+# but it does not work. install pytables from pypi.
```
mtbc, line 9: :+1: for explanatory comment

joshmoore:
Changes make sense, however, I do see:

```
travis_time:start:1548f5e3
[0K$ ./autogenerate.sh
centos7
centos6_py27
sed: -e expression #1, char 4: invalid usage of line address 0
sed: -e expression #1, char 4: invalid usage of line address 0
centos6_py27_ius
ubuntu1404
debian8
ubuntu1604
debian9
```

in the travis log.

As a side question: does the fact that debian9 was in the travis matrix before this PR but there was no failure point to an issue with the test(s)?

jburel:
The error seems strange. I do not have it locally

jburel:
It does not seem related to those changes
I can see a similar output on https://github.com/ome/omero-install/pull/165

jburel:
It looks like it started in https://github.com/ome/omero-install/pull/162

mtbc:
I think that those sed errors are because `autogenerate.sh` expects `step05_centos6_py27_nginx.sh` to have the `nginx-install` and `nginx-admin` comments.

jburel:
travis log issue fixed in https://github.com/ome/omero-install/pull/170

joshmoore:
:+1: for gh-170. Anything on 

> As a side question: does the fact that debian9 was in the travis matrix before this PR but there was no failure point to an issue with the test(s)?

?

jburel:
We will probably need to parse the output of autogen
(not related to debian 9) since it should have failed
Nothing to urgent on that front.

PR #172, opened 2017-11-17
gunicorn debian

jburel:
Install gunicorn using package manager

https://trello.com/c/21YowfLZ/109-clarify-gunicorn-installation-for-omeroweb-installation-with-server-on-debian

Tested via travis


mtbc:
For my Debian 9 OpenStack images I indeed use `apt-get install gunicorn`.

PR #173, opened 2018-01-12
Set omero systemd UMask=0002

manics:
Set the default OMERO systemd umask to 0022 to enable in-place imports to work by enabling group-write. PR review should include a consideration of whether this relaxation of permissions is justified. If it's not we should update https://github.com/openmicroscopy/ansible-role-omero-server/blob/2.0.2/defaults/main.yml#L28 instead

See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2018-January/006865.html

### DIFF
--- a/linux/omero-systemd.service
+++ b/linux/omero-systemd.service
@@ -9,6 +9,7 @@ After=network.service
 
 [Service]
 User=omero
+UMask=0002
 Type=forking
 PIDFile=/home/omero/OMERO.server/var/master/master.pid
 Restart=no
--- a/linux/omero-systemd.service
+++ b/linux/omero-systemd.service
@@ -9,7 +9,6 @@ After=network.service
 
 [Service]
 User=omero
-UMask=0002
 Type=forking
 PIDFile=/home/omero/OMERO.server/var/master/master.pid
 Restart=no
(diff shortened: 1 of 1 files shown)


joshmoore:
Restarted https://travis-ci.org/ome/omero-install/jobs/328065516

joshmoore:
> review should include a consideration of whether this relaxation of permissions is justified

I'm not sure I know enough about what was being attempted. Is there a description of the problem(s) somewhere?

manics:
Our in-place import docs say a umask of `0002` is required https://docs.openmicroscopy.org/omero/5.4.1/sysadmins/in-place-import.html

Is this an acceptable default for all OMERO servers?

(Also there's a typo in this PR..I'll force push)

joshmoore:
I don't think we specify anywhere assumptions about the `omero(-server)` user's group, therefore assuming that `0002` is safe on all servers is probably an issue. Without digging deeper, I'd err on the side of making that an addition to the in-place documentation, i.e. a sysadmin is knowingly opening up permissions on a system because of a level of trust granted to the in-place group.

manics:
Updated

joshmoore:
Seems like an innocuous enough comment. Merging in time for 5.4.2.

Issue #112, opened 2016-06-17
ice36-devel is missing


joshmoore:
It's slightly confusing that ice36 installs devel requirements (like slice2py) but ice35 doesn't. Probably worth unifying. (Note: this came from trying to change ice36 in devspace to ice35 and then having slice2java not on the path)

atarkowska:
https://github.com/ome/omero-install/blob/develop/linux/step01_centos7_ice_deps.sh

manics:
Does ice36 need the devel packages at all?

joshmoore:
@manics : quite possibly not, and it might be just that disparity that leads to confusion.

joshmoore:
@jburel: worth reviewing for 5.3?

jburel:
will try to see if I can have something today otherwise will look at it in 5.3.1

jburel:
unlikely now, I had to work on homebrew

jburel:
I came back to that issue today
It is not possible to install the python binding i.e. ``pip install "zeroc-ice>3.5,<3.7"``
if ``openssl-devel`` and  ``bzip2-devel`` are not installed
``expat-devel`` is not necessary. This can be removed.

jburel:
https://github.com/ome/omero-install/pull/175

PR #175, opened 2018-02-10
ice 36 centos7

jburel:
Fix ice installation
and introduce ice36-devel
see https://github.com/ome/omero-install/issues/112


joshmoore (commented):
Barring the difference in whitespace (tabs v. spaces), this looks ok.

 * :+1: drop expat-devel in both
 * :+1: drop ice-all-devel in ice36
 * :+1: add ice-all-devel in ice36-devel (only difference to ice36)

Issue #168, opened 2017-09-18
PostgreSQL on Ubuntu requires configuration


joshmoore:
See https://github.com/openmicroscopy/openmicroscopy/issues/5508

joshmoore:
cF. https://github.com/ome/omero-install/blob/develop/linux/step01_centos7_pg_deps.sh#L17

jburel:
see https://github.com/ome/omero-install/pull/176

PR #176, opened 2018-02-12
Postgresql configuration

jburel:
Postgresql configuration ubuntu

See https://github.com/ome/omero-install/issues/168

Check that there is no error in log

### DIFF
--- a/linux/step01_ubuntu1604_pg_deps.sh
+++ b/linux/step01_ubuntu1604_pg_deps.sh
@@ -7,6 +7,7 @@ PGVER=${PGVER:-pg94}
 if [ "$PGVER" = "pg95" ]; then
 	apt-get update
 	apt-get -y install postgresql
+	sed -i.bak -re 's/^(host.*)ident/\1md5/' /etc/postgresql/9.5/main/pg_hba.conf
 	service postgresql start
 elif [ "$PGVER" = "pg94" ]; then
 	apt-get -y install apt-transport-https
@@ -14,6 +15,7 @@ elif [ "$PGVER" = "pg94" ]; then
 	wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
 	apt-get update
 	apt-get -y install postgresql-9.4
+	sed -i.bak -re 's/^(host.*)ident/\1md5/' /etc/postgresql/9.4/main/pg_hba.conf
 	service postgresql start
 elif [ "$PGVER" = "pg96" ]; then
 	#start-recommended
@@ -22,6 +24,7 @@ elif [ "$PGVER" = "pg96" ]; then
 	wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
 	apt-get update
 	apt-get -y install postgresql-9.6
+	sed -i.bak -re 's/^(host.*)ident/\1md5/' /etc/postgresql/9.6/main/pg_hba.conf
 	service postgresql start
 	#end-recommended
 fi


waxenegger:
I can confirm that it works. No errors on script run, ident has been all replaced with md5, and restarting omero everything works fine, connecting to the database with psql works showing that an SSL connection was used .

```root@5bdee890681a:/opt/omero-install# bash linux/step01_ubuntu1604_pg_deps.sh 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
apt-transport-https is already the newest version (1.2.25).
0 upgraded, 0 newly installed, 0 to remove and 8 not upgraded.
OK
Hit:1 http://security.ubuntu.com/ubuntu xenial-security InRelease                                                                     
Hit:3 http://archive.ubuntu.com/ubuntu xenial InRelease                                                                               
Hit:4 http://ppa.launchpad.net/openjdk-r/ppa/ubuntu xenial InRelease                          
Hit:5 http://archive.ubuntu.com/ubuntu xenial-updates InRelease                               
Hit:6 http://archive.ubuntu.com/ubuntu xenial-backports InRelease                    
Hit:2 http://zeroc.com/download/apt/ubuntu16.04 stable InRelease
Hit:7 https://apt.postgresql.org/pub/repos/apt xenial-pgdg InRelease
Reading package lists... Done
Reading package lists... Done
Building dependency tree       
Reading state information... Done
postgresql-9.4 is already the newest version (9.4.16-1.pgdg16.04+1).
0 upgraded, 0 newly installed, 0 to remove and 8 not upgraded.
 * Starting PostgreSQL 9.4 database server                                                                                      [ OK ] 
root@5bdee890681a:/opt/omero-install# cat /etc/postgresql/9.4/main/pg_hba.conf
# PostgreSQL Client Authentication Configuration File
# ===================================================
#
# Refer to the "Client Authentication" section in the PostgreSQL
# documentation for a complete description of this file.  A short
# synopsis follows.
#
# This file controls: which hosts are allowed to connect, how clients
# are authenticated, which PostgreSQL user names they can use, which
# databases they can access.  Records take one of these forms:
#
# local      DATABASE  USER  METHOD  [OPTIONS]
# host       DATABASE  USER  ADDRESS  METHOD  [OPTIONS]
# hostssl    DATABASE  USER  ADDRESS  METHOD  [OPTIONS]
# hostnossl  DATABASE  USER  ADDRESS  METHOD  [OPTIONS]
#
# (The uppercase items must be replaced by actual values.)
#
# The first field is the connection type: "local" is a Unix-domain
# socket, "host" is either a plain or SSL-encrypted TCP/IP socket,
# "hostssl" is an SSL-encrypted TCP/IP socket, and "hostnossl" is a
# plain TCP/IP socket.
#
# DATABASE can be "all", "sameuser", "samerole", "replication", a
# database name, or a comma-separated list thereof. The "all"
# keyword does not match "replication". Access to replication
# must be enabled in a separate record (see example below).
#
# USER can be "all", a user name, a group name prefixed with "+", or a
# comma-separated list thereof.  In both the DATABASE and USER fields
# you can also write a file name prefixed with "@" to include names
# from a separate file.
#
# ADDRESS specifies the set of hosts the record matches.  It can be a
# host name, or it is made up of an IP address and a CIDR mask that is
# an integer (between 0 and 32 (IPv4) or 128 (IPv6) inclusive) that
# specifies the number of significant bits in the mask.  A host name
# that starts with a dot (.) matches a suffix of the actual host name.
# Alternatively, you can write an IP address and netmask in separate
# columns to specify the set of hosts.  Instead of a CIDR-address, you
# can write "samehost" to match any of the server's own IP addresses,
# or "samenet" to match any address in any subnet that the server is
# directly connected to.
#
# METHOD can be "trust", "reject", "md5", "password", "gss", "sspi",
# "ident", "peer", "pam", "ldap", "radius" or "cert".  Note that
# "password" sends passwords in clear text; "md5" is preferred since
# it sends encrypted passwords.
#
# OPTIONS are a set of options for the authentication in the format
# NAME=VALUE.  The available options depend on the different
# authentication methods -- refer to the "Client Authentication"
# section in the documentation for a list of which options are
# available for which authentication methods.
#
# Database and user names containing spaces, commas, quotes and other
# special characters must be quoted.  Quoting one of the keywords
# "all", "sameuser", "samerole" or "replication" makes the name lose
# its special character, and just match a database or username with
# that name.
#
# This file is read on server startup and when the postmaster receives
# a SIGHUP signal.  If you edit the file on a running system, you have
# to SIGHUP the postmaster for the changes to take effect.  You can
# use "pg_ctl reload" to do that.

# Put your actual configuration here
# ----------------------------------
#
# If you want to allow non-local connections, you need to add more
# "host" records.  In that case you will also need to make PostgreSQL
# listen on a non-local interface via the listen_addresses
# configuration parameter, or via the -i or -h command line switches.




# DO NOT DISABLE!
# If you change this first entry you will need to make sure that the
# database superuser can access the database using some other method.
# Noninteractive access to all databases is required during automatic
# maintenance (custom daily cronjobs, replication, and similar tasks).
#
# Database administrative login by Unix domain socket
local   all             postgres                                peer

# TYPE  DATABASE        USER            ADDRESS                 METHOD

# "local" is for Unix domain socket connections only
local   all             all                                     peer
# IPv4 local connections:
host    all             all             127.0.0.1/32            md5
# IPv6 local connections:
host    all             all             ::1/128                 md5
# Allow replication connections from localhost, by a user with the
# replication privilege.
#local   replication     postgres                                peer
#host    replication     postgres        127.0.0.1/32            md5
#host    replication     postgres        ::1/128                 md5
root@5bdee890681a:/opt/omero-install# service omero restart
Stopping omero: OMERO.server
Starting omero: OMERO.server
root@5bdee890681a:/opt/omero-install# psql -h localhost -U db_user -d omero_database
Password for user db_user: 
psql (9.4.16)
SSL connection (protocol: TLSv1.2, cipher: ECDHE-RSA-AES256-GCM-SHA384, bits: 256, compression: off)
Type "help" for help.

omero_database=> 
```

PR #178, opened 2018-02-13
Remove ice-devel ubuntu

jburel:
Remove ice-dev package

Check that travis is green
Locally build ubuntu16.04 and start the server and try to connect via web and insight


rleigh-codelibre:
Looking at https://travis-ci.org/ome/omero-install/jobs/340908548 this seems safe; it doesn't break the pip install of the ice python bindings.

Not sure how I would test it locally.  Are there any instructions?

jburel:
Instructions are in README
https://github.com/ome/omero-install/blob/develop/linux/test/README.md

Run first ``./docker-build.sh ubuntu1604_nginx``
Then 
``docker run --rm -it -p 8080:80 -p 4063:4063 -p 4064:4064 omero_install_test_ubuntu1604_nginx``

Check that you can import an image and view it in Web 
will be enough

rleigh-codelibre:
Hmm, I'll have to try it on beluga.

rleigh-codelibre:
I can log in with insight but the web server port is immediately closing the connection.  Web client login page shows but there's some problem with the mapped ports I think--it won't let me log in.  Looks fine to me and good to merge.

PR #179, opened 2018-02-19
Clarification

jburel:
Indicate to enter a password for the omero user

cc @rgozim


rgozim:
Everything working fine. Setting password for the _omero_ user allows me to logout of my current user and log back in again as the new account.

PR #182, opened 2018-08-03
Remove pip upgrade

jburel:
Recent upgrade of pip broke the installation
This PR removes the upgrade since it is not required.
See https://trello.com/c/BlTTfNgW/16-omero-server-installation-issues-centos7-pip

and also cap version of packages installed using pip

check that travis is green


sbesson:
Is there a baseline commit reproducing the CentOS7 issues with the pip upgrade? The logs of https://travis-ci.org/ome/omero-install/jobs/408409981 do not indicate any mention of `pip install --upgrade pip` which should have failed the build as per the card. Does that imply that part of the code copied to the documentation was effectively not executed?

jburel:
For centos7 it was not executed when we run it in docker

rgozim:
Looks okay to me. I'm going to assume there's probably not much need in upgrading pip beyond what is supplied by the platform packages for our use cases.

jburel:
Actually no need to upgrade pip
I have also capped the version of  packages installed via pip since one was leading to a failure (C6 Py27)

joshmoore:
Changes look generally fine:
 * I assume previously `pip install --upgrade pip` wasn't harmful and therefore we didn't realize.
 * I'm not clear on where the pinned virtualenv version comes from.
 * You assume the next steps would be to close https://github.com/openmicroscopy/ome-documentation/pull/1890 and then re-generate?

> For centos7 it was not executed when we run it in docker

Why was this?

jburel:
* It was not harmful until we reach newer version of pip.
* The ``virtualenv`` cap is there to fix the failure we have C6 i.e. similar to pip upgrade. We did not have a problem until we reach a version of ``virtualenv``
* We had a ``if docker`` but I cannot remember we had it.

I will close the doc PR. When this is merged, the doc will be updated after running autogen

joshmoore:
Ok merging so that we can see what is produced via autogen.

PR #186, opened 2018-12-10
Update ice 3.6 install

jburel:
Install ice py using wheel

This will speed the installation
Check that travis is green and you can start the server
i.e.
``
./docker-build.sh ubuntu1604_nginx
 docker run --rm -it -p 8080:80 -p 4063:4063 -p 4064:4064 omero_install_test_ubuntu1604_nginx
``

centOS 7 file will be updated after it is moved to ``ome``


dominikl:
It seems to work fine. Can run the docker and OMERO.web is availabe on 'http://<docker_machin_ip>:8080/webclient/' but when I try to log in I get
"Error: Server is not responding, please contact administrator". Also can't login using Insight. But I assume that's just a docker portforwarding issue or something like that, because everything looks ok and I can't even see a login attempt in the Blitz.log.

jburel:
Why credentials did you use? I can show you tomorrow

dominikl:
Works, good to merge 👍 
Turned out that my docker vm just didn't have enough memory.

PR #189, opened 2018-12-12
Remove Variable 

jburel:
Remove variable no longer needed
This was required for Python 2.7 support on CentOS 6

### DIFF
--- a/linux/step05_1_all_webapps.sh
+++ b/linux/step05_1_all_webapps.sh
@@ -43,7 +43,7 @@ if [ $install_figure = true ]; then
 	# Install required packages
 	if [ "$PY_ENV" = "py27" ]; then
 		pip install reportlab markdown
-	else
+	fi
 
 	# Register the app
 	su - omero -c "OMERO.server/bin/omero config append omero.web.apps '\"figure\"'"
--- a/linux/step05_1_all_webapps.sh
+++ b/linux/step05_1_all_webapps.sh
@@ -40,26 +40,10 @@ if [ $install_figure = true ]; then
 	rm $NAME_FIGURE_ZIP
 	mv figure* OMERO.server/lib/python/omeroweb/figure
 
-	echo "value=$PY_ENV"
 	# Install required packages
-	if [ "$PY_ENV" = "py26" ]; then
-		pip install reportlab==2.7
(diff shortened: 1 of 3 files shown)


joshmoore:
Will a similar variable be needed for PY3? (Not that it needs adding now; I'm just wondering if this will need adding back in soon)

jburel:
Maybe so I will revisit the cleaning and if it is not required
when we work on PY3  I can do a full cleaning

linux/step05_1_all_webapps.sh
```diff
@@ -40,26 +40,10 @@ if [ $install_figure = true ]; then
 	rm $NAME_FIGURE_ZIP
 	mv figure* OMERO.server/lib/python/omeroweb/figure
 
-	echo "value=$PY_ENV"
 	# Install required packages
```
mtbc: Is the comment still needed?

linux/step05_1_all_webapps.sh
```diff
@@ -40,26 +40,10 @@ if [ $install_figure = true ]; then
...
-		set -u
-		pip install reportlab markdown
-	elif [ "$PY_ENV" = "py27_ius" ]; then
-		virtualenv -p /usr/bin/python2.7 /home/omero/omeroenv
-		set +u
-		source /home/omero/omeroenv/bin/activate
-		set -u
-		/home/omero/omeroenv/bin/pip2.7 install reportlab markdown
-		deactivate
 	else
```
mtbc: Should this be `fi`?

linux/step05_2_websessionconfig.sh
```diff
@@ -4,18 +4,7 @@ PY_ENV=${PY_ENV:-py27}
 
 cd ~omero
 
-echo "value=$PY_ENV"
 # Install required packages
```
mtbc: Is this comment still needed?

jburel:
Thanks @mtbc 
It should be fixed now

mtbc:
:+1: I don't see any other problems.

Outwith this PR, in the longer term we might want more `else`s that alert of unset/unexpected values.

jburel:
This is mainly b/c that code is not really used. I might get rid of it since we install the apps differently now

PR #190, opened 2018-12-17
Debian 9 install

jburel:
Use wheel to install python binding

Check that travis is green

### DIFF
--- a/linux/step01_debian9_ice_deps.sh
+++ b/linux/step01_debian9_ice_deps.sh
@@ -5,12 +5,8 @@ ICEVER=${ICEVER:-ice36}
 
 # Ice installation
 if [ "$ICEVER" = "ice36" ]; then
-	#start-recommended
-	# to be installed if recommended/suggested is false
-	apt-get -y install build-essential python-dev
-
- 	apt-get -y install libssl-dev libbz2-dev libmcpp-dev libdb++-dev libdb-dev libdb-java
+    #start-recommended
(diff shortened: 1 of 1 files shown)


mtbc:
Diff certainly looks plausible. Curious as to why the replaced `pip install` approach is bad (as it's what I do on OpenStack).

joshmoore:
This method is just much faster since the wheel is pre-built.

dominikl:
Trying to test the PR locally with `./docker-build.sh debian9_nginx` but get:
```
+ echo 'CREATE USER db_user PASSWORD '\''db_password'\'''
No passwd entry for user 'postgres'
The command '/bin/sh -c cd omero-install-test && 	bash install_debian9_nginx.sh && 	bash docker_shutdown_ubuntu.sh' returned a non-zero code: 1
```
Although looks to be unrelated to the PR.

mtbc:
Huh, maybe yet another Docker weirdness? Might be worth trying OpenStack instead.

jburel:
@dominikl thanks I don't think it is a docker weirdness but it is more related to the version of pg used by default in that case. 
Let me review that. This is outside the scope of this PR
If you run ``PGVER=pg96 ./docker-build.sh debian9_nginx`` it will work 

I will fix it in another PR

jburel:
Problem mentioned by @dominikl is not related to this PR and it has now been fixed
see https://github.com/ome/omero-install/pull/191

jburel:
Merging

PR #197, opened 2019-05-16
Debian9: Use omero.glacier2.IceSSL.Ciphers to set @SECLEVEL=0

manics:
Use the [new `omero.glacier2.IceSSL.Ciphers` property](https://github.com/openmicroscopy/openmicroscopy/pull/5998) on https://docs.openmicroscopy.org/omero/5.4.10/sysadmins/troubleshooting.html#openssl-version

~Also use `sed -i` for the remaining file to prevent the `.bak` file being created.~

~If https://github.com/openmicroscopy/openmicroscopy/pull/6007 is judged safe then the `sed ... clients.py` command could be removed completely.~ This was merged

PR #198, opened 2019-05-17
test_services.sh should work for all distros

manics:
`./test_services.sh` is now run for all distros, not just centos-7. This requires changes to the `run.sh` script to ensure the container remains alive in non-interactive (daemon) mode.

Also removed the out-of-date Darwin specific conditionals.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,6 @@ before_install:
 script:
     - cd linux/test
     - PGVER=$PGVER ./docker-build.sh $ENV
-    - if [[ "$ENV" =~ "centos7" ]]; then ./test_services.sh; fi
+    - ./test_services.sh
     - cd ..
     - ./autogenerate.sh
--- a/linux/test/debian9_nginx/run.sh
+++ b/linux/test/debian9_nginx/run.sh
@@ -26,4 +26,8 @@ cron
 service omero start
 service nginx start
 service omero-web start
-exec tail -F /home/omero/OMERO.server/var/log/*
+if [ -t 1 ] ; then
+    exec bash
+else
+    exec tail -F /home/omero/OMERO.server/var/log/*
+fi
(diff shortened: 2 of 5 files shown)


manics:
Be aware that if https://github.com/ome/omero-install/pull/197 is merged this will start failing since it's using the last release

jburel:
This certainly allows to test via travis but it prevents one from using the image locally  in a usable manner: i.e.
``docker run --rm -it -p 8080:80 -p 4063:4063 -p 4064:4064 omero_install_test_ubuntu1604_nginx``

joshmoore:
What's broken for the local usage, @jburel ?

jburel:
It was useful to be able to access the image and run other command for example now you have the log in the terminal which is not very practical.

joshmoore:
Trying to pinpoint what you mean precisely?

> exec tail -F /home/omero/OMERO.server/var/log/*

?

jburel:
yes replacing ``bash`` by ``tail...`` is the source of that broken workflow

joshmoore:
Understood. Then I'd propose:

```
bash-3.2$ docker run --rm josh-tmp
not a terminal
tail: cannot open '/var/log/messages' for reading: No such file or directory
tail: no files remaining
bash-3.2$ docker run --rm -ti josh-tmp
terminal
[root@5e79acb56cf7 /]#
```

```
==> Dockerfile <==
FROM centos:centos7
ADD entry.sh /entry.sh
CMD ["/entry.sh"]

==> entry.sh <==
#!/usr/bin/env bash
if [ -t 1 ] ; then
    echo terminal
    exec bash
else
    echo "not a terminal"
    tail -f /var/log/messages
fi
```

jburel:
I have pushed a commit including the suggestion. This allows support for both workflows

jburel:
Merging

PR #199, opened 2019-05-17
Quieten download progress output

manics:
Removes the excessive progress output when downloading OMERO.


jburel:
Tested locally. Download works fine and it is quieter. Thanks for spotting it.
Merging

PR #201, opened 2019-05-23
cleaning

jburel:
remove extra bracket

Warning noticed while working on ubuntu 18.04
Check that travis is green


joshmoore:
:+1: but it's a bit odd that it was just a warning and not an error, no?

jburel:
that's the output without the change
```
+ '[' py27 = py27 ']]'
step04_all_omero.sh: line 12: [: missing `]'
```
and the script continues

joshmoore:
:+1:. The fact that the script continues is likely something we should look into.

PR #202, opened 2019-05-24
latest 5.5

jburel:
point to omero5.5 (5.5.0-rc1)
This is required otherwise SECLEVEL change done https://github.com/ome/omero-install/pull/197
will imply that build will fail

The server currently does not start on Java 11. I discovered that last night. disabling for now the java 11 jobs
cc @mtbc


jburel:
Merging so new PR will pas

PR #206, opened 2019-06-21
Correct name of PostgreSQL setup script.

mtbc:
```
$ cat /etc/centos-release
CentOS Linux release 7.6.1810 (Core) 
$ ls -l /usr/pgsql-10/bin/postgresql*setup
-rwxr-xr-x. 1 root root 9419 May  8 10:25 /usr/pgsql-10/bin/postgresql-10-setup
$ 
```

### DIFF
--- a/linux/step01_centos7_pg_deps.sh
+++ b/linux/step01_centos7_pg_deps.sh
@@ -78,7 +78,7 @@ elif [ "$PGVER" = "pg10" ]; then
         su - postgres -c "/usr/pgsql-10/bin/initdb -D /var/lib/pgsql/10/data --encoding=UTF8"
         echo "listen_addresses='*'" >> /var/lib/pgsql/10/data/postgresql.conf
     else
-        PGSETUP_INITDB_OPTIONS=--encoding=UTF8 /usr/pgsql-10/bin/postgresql10-setup initdb
+        PGSETUP_INITDB_OPTIONS=--encoding=UTF8 /usr/pgsql-10/bin/postgresql-10-setup initdb
     fi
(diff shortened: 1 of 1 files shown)


jburel:
Strange: Travis has always been green and PG10 is the testing matrix
see https://github.com/ome/omero-install/blob/develop/.travis.yml#L5

joshmoore:
Feels vaguely similar to the recent issue where a bash script was failing silently. I imagine adding `set -e` to the proper locations would detect this type of thing.

mtbc:
No idea what the Travis failures are about.

joshmoore:
Looks like it never found `OMERO.blitz now accepting connections` in quite a few of the servers. Perhaps that script could cat the Blitz-0.log file on failure?

jburel:
Having a closer look at the file, the reason travis did not fail is b/c the command is not run in docker!
that's the commands run if docker container
```
su - postgres -c "/usr/pgsql-10/bin/initdb -D /var/lib/pgsql/10/data --encoding=UTF8"
echo "listen_addresses='*'" >> /var/lib/pgsql/10/data/postgresql.conf
```

jburel:
Thanks for spotting that error

joshmoore:
Is this a good target for trying to pin the postgres version?

mtbc:
(I don't know but happy to accept whatever you and @jburel think wise.)

jburel:
I think we can merge this PR as it is for 5.5.1
The change is not tested in travis and it will correct our doc.
Let's hope a new pg version will be released in the coming days. Otherwise we will have to pin the version (this will require another doc release)

joshmoore:
Merging along with https://github.com/ome/omero-install/pull/208#issuecomment-507703522

PR #207, opened 2019-06-24
remove warning

jburel:
delete zip to avoid msg 'binary operator expected'
when checking if the server exists


joshmoore:
Merging also without a functioning travis (see https://github.com/ome/omero-install/pull/208#issuecomment-507703522). We will need to review the walk-throughs manually.

cc: @ome/release

PR #208, opened 2019-06-26
CentOS 7 use wheel

jburel:
Speed up the installation of ice py
This matches what is currently done for Ubuntu and Debian


joshmoore:
:+1: though now I'm really starting to wonder if we should maintain a redirect for this.

joshmoore:
(Failure is likely the postgres issue)

jburel:
It is due to pg issue

joshmoore:
I tried to find a way to fix PG but to no avail: https://twitter.com/notjustmoore/status/1146040843607904256

Merging to see the effects on autogen.

PR #211, opened 2019-07-02
Pin postgresql versions

joshmoore:
see:
 * https://forum.image.sc/t/fyi-not-possible-to-create-new-omero-instance-with-latest-postgresql-versions/26929
 * https://github.com/ome/ansible-role-postgresql/pull/15

### DIFF
--- a/linux/step01_centos7_pg_deps.sh
+++ b/linux/step01_centos7_pg_deps.sh
@@ -6,7 +6,7 @@ PGVER=${PGVER:-pg94}
 if [ "$PGVER" = "pg94" ]; then
 	# Postgres, reconfigure to allow TCP connections
 	yum -y install http://yum.postgresql.org/9.4/redhat/rhel-7-x86_64/pgdg-redhat94-9.4-3.noarch.rpm
-	yum -y install postgresql94-server-9.4.22 postgresql94-9.4.22
+	yum -y install postgresql94-server postgresql94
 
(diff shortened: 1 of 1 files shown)


joshmoore:
This is in need of a postgres fix for ubuntu. Various things I tried before asking on twitter:

 * I don’t see debs for the older versions including with http://apt.postgresql.org/pub/repos/apt/ (https://wiki.postgresql.org/wiki/Apt)
 * http://archive.ubuntu.com/ubuntu/pool/main/p/postgresql-9.6/ is empty
 * http://archive.ubuntu.com/ubuntu/pool/main/p/postgresql-10/ has 10.5 but not 10.8
 * well, https://packages.debian.org/stretch/amd64/postgresql-9.6/download has it!
 * jessie has 0.4, buster/sid have 11, but no 9.5 or 10 :disappointed:
 * Is stuff from http://apt.postgresql.org/pub/repos/apt/pool/ any use?  No, the pool is missing previous versions.
 * https://snapshot.debian.org/ perhaps? They don’t include the patch version though. https://snapshot.debian.org/package/postgresql-common/184/#postgresql_9.6:2b:184 ...  and they’re all from 2017?!

joshmoore:
Debian and Ubuntu versions are failing, not CentOS so I _think_ this best idea is to just close this and we deal with those issues elsewhere.

PR #213, opened 2019-07-10
useradd system user

manics:
Create the `omero` user as a system account.
See https://forum.image.sc/t/release-of-omero-5-5-0/26600/8

### DIFF
--- a/linux/step02_all_setup.sh
+++ b/linux/step02_all_setup.sh
@@ -2,7 +2,7 @@
 
 if [ -z "$(getent passwd omero)" ]; then
 	#start-create-user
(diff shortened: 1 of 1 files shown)


dominikl:
Seems to work, uid < 1000 == system user 👍 
```
$ getent passwd
...
omero:x:996:993::/home/omero:/bin/bash
...
```

Although I can't start omero `org.postgresql.util.PSQLException: ERROR: relation "dbpatch" does not exist` but that is expected apparently and not related to this PR.

jburel:
cherry-picked the commit and added to #216

PR #218, opened 2019-12-04
Server location

jburel:
Change the location where OMERO.server is installed
This is now under /opt/omero/server/OMERO.server
Download is now done by root, then the directory is owned by the omero system user.

cc @joshmoore @manics


joshmoore:
Merging along with https://github.com/ome/ome-documentation/pull/2044

PR #221, opened 2019-12-06
Remove web steps

jburel:
Remove web sections since we no longer use any of it
Web installation is at https://github.com/ome/omeroweb-install

Check that travis is green

cc @dominikl


dominikl:
Looks good. Tested the centos7 and debian9 installation, works fine.
Well, debian9 after some additional steps. Not related to the PR, but maybe we could add
```
sudo systemctl disable glacier2router
sudo systemctl disable icegridregistry
```
somewhere in the debian installation script, otherwise omero-server won't start up because the 'address is already in use'.

jburel:
it is there already in the doc See https://docs.openmicroscopy.org/omero/5.6.0-m4/sysadmins/unix/server-debian9-ice36.html

dominikl:
Strange... sure this does happen when you run the `install_debian9.sh`? I had to manually do that afterwards. But anyway, not a matter of this PR.

jburel:
Are you using docker?
The command is not part of omero-install. I could add a condition ``if not docker do X``
and include it in the autogen doc

dominikl:
No, I used a debian9 VM and ran the `linux/install_debian9.sh` script.

jburel:
ok so I will add an extra step

PR #222, opened 2019-12-11
install wheel

jburel:
Make sure python3-wheel is in the dependencies list

background https://forum.image.sc/t/omero-python-3-5-6-milestone-for-upgrade-testing/32120/8


dominikl:
Works 👍 Strange that this is only needed for ubuntu and not for debian.

jburel:
Actually looking at the log it did not have the expected effect.
Rolling back the change

PR #225, opened 2019-12-12
Wheel rollback

jburel:
Looking closely at the log, I could see ``invalid command 'bdist_wheel'``
I have installed ``wheel`` in the virtual env and the error is no longer there

* Check that travis is green
* Check that the error ``invalid command 'bdist_wheel'`` is not in the travis log

cc @dominikl


jburel:
Closing, replaced by https://github.com/ome/omero-py/pull/147

PR #226, opened 2019-12-20
settings

jburel:
re-organise for doc


jburel:
Green merging so it can be fixed in the doc

PR #216, opened 2019-11-25
Py3 setup

jburel:
* Update to use python3 and virtual env
* For convenience, we only use one virtualenv. Web is kept for convenience but it will no longer appears in the ome-doc i.e. we recommend server and web to be separated
Web installation doc is generated by omeroweb-install

PR #230, opened 2020-01-22
install pytables

jburel:
Pytables were previously installed using the package manager
This was removed in https://github.com/ome/omero-install/pull/216
under the assumption that this will be included in ``omero-py``
this is not the case.
This should fix the missing installation step in our guide
See https://github.com/ome/ome-documentation/pull/2097

### DIFF
--- a/linux/step01_centos7_ice_venv.sh
+++ b/linux/step01_centos7_ice_venv.sh
@@ -9,4 +9,7 @@ python3 -mvenv $VENV_SERVER
 
 # Install the Ice Python binding
 $VENV_SERVER/bin/pip install https://github.com/ome/zeroc-ice-py-centos7/releases/download/0.2.1/zeroc_ice-3.6.5-cp36-cp36m-linux_x86_64.whl
+
+# Install pytables
+$VENV_SERVER/bin/pip install tables
 #end-ice-py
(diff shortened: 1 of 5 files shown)


joshmoore:
The alternative would be to add to omero-py's setup.py , right?

jburel:
yes

jburel:
I will have to work on it a bit I remember putting a workaround for debian9 ubuntu16.04

jburel:
Depending on the OS version we need to pin pytables so it might be better not to add it to omero-py.
But that will also means that we will need to review the conda step for "omero-server"

will-moore:
If you're using ```omero-py``` purely as a client, then you don't need ```tables```, so it probably shouldn't be mandatory?

manics:
I agree, let's not bundle every possible dependency into omero-py setup.py

pwalczysko:
I think we should progress here - pytables are crucial to have on outreach servers for the next week workshop.

manics:
pytables is already installed on outreach:
```
$ /opt/omero/server/venv3/bin/pip list
Package             Version
------------------- -------
cycler              0.10.0
future              0.18.2
Jinja2              2.10.3
kiwisolver          1.1.0
Markdown            3.1.1
MarkupSafe          1.1.1
matplotlib          3.1.2
numexpr             2.7.1
numpy               1.18.1
omego               0.7.0
omero-cli-duplicate 0.3.0
omero-cli-render    0.5.0
omero-dropbox       5.6.1
omero-metadata      0.5.0
omero-py            5.6.0
Pillow              7.0.0
pip                 18.1
pyparsing           2.4.6
python-dateutil     2.8.1
PyYAML              5.3
reportlab           3.5.34
scipy               1.4.1
setuptools          40.6.2
six                 1.14.0
tables              3.6.1
yaclifw             0.2.0
zeroc-ice           3.6.5
```

will-moore:
Looks good 👍

PR #232, opened 2020-03-22
include PATH in systemd.service

jburel:
Matching PR 
https://github.com/ome/omeroweb-install/pull/58

see https://forum.image.sc/t/omero-web-systemd-file-is-missing-path/35248/3


joshmoore:
General thumbs up during chat today. Merging for the 5.6.1 build.

Issue #234, opened 2020-04-20
PG URLs failling on travis


joshmoore:
https://travis-ci.org/github/ome/omero-install/jobs/676151531 is failing with:

```
+ yum -y install https://yum.postgresql.org/10/redhat/rhel-7-x86_64/pgdg-redhat10-10-2.noarch.rpm
1318Loaded plugins: fastestmirror, ovl
1319Cannot open: https://yum.postgresql.org/10/redhat/rhel-7-x86_64/pgdg-redhat10-10-2.noarch.rpm. Skipping.
1320Error: Nothing to do
```

similar to https://travis-ci.org/github/ome/omero-install/jobs/676151536 :

```
+ yum -y install https://yum.postgresql.org/11/redhat/rhel-7-x86_64/pgdg-redhat11-11-2.noarch.rpm
1336Loaded plugins: fastestmirror, ovl
1337Cannot open: https://yum.postgresql.org/11/redhat/rhel-7-x86_64/pgdg-redhat11-11-2.noarch.rpm. Skipping.
1338Error: Nothing to do
```

sbesson:
See also https://github.com/ome/ansible-role-postgresql/pull/17 updated the `ome.postgresql` role to use the new style repo RPMs. It is possible the old RPMs have now been dropped

jburel:
see https://github.com/ome/omero-install/pull/235

PR #235, opened 2020-04-23
update pg url

jburel:
This PR updates the url. The job should turn back to green 
see https://github.com/ome/omero-install/issues/234

When merged, we will probably need a doc release

Check that travis is green

### DIFF
--- a/linux/step01_centos7_pg_deps.sh
+++ b/linux/step01_centos7_pg_deps.sh
@@ -70,7 +70,7 @@ elif [ "$PGVER" = "pg96" ]; then
 	fi
 	systemctl enable postgresql-9.6.service
 elif [ "$PGVER" = "pg10" ]; then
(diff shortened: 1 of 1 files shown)


sbesson:
Travis CI is back to green and the proposed change make sense. Two minor questions:

- since the `yum install` command is now independent of the PSQL version, is it feasible to move it before each conditional block?
- also is it worth switching Travis to test PSQL11 on Centos7 as this is the recommended version?

jburel:
* Travis already tests pg11 by default on centos, debian and ubuntu, it is the default value in the file i.e. ``PGVER=${PGVER:-pg11}``
* the url is different from one version to the other i.e. the version number is in the url

sbesson:
>the url is different from one version to the other i.e. the version number is in the url

Of course. I should not review PRs as I am waking up...

sbesson (approved):

PR #237, opened 2020-04-29
Use omero-certificates

jburel:
install and configure omero-certificates

### DIFF
--- a/linux/autogenerate.sh
+++ b/linux/autogenerate.sh
@@ -70,8 +70,6 @@ echo "# install dependencies" >> $file
 N=$OS
 if [[ $OS =~ "ubuntu" ]] ; then
 	N="ubuntu1604"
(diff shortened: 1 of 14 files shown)


joshmoore:
The related section is no longer "patching OMERO.server":

Screen Shot 2020-05-06 at 13 33 01

jburel:
I have not really worked on the doc only re-use the file used to "patch", that's why it shows up like this

joshmoore:
> I have not really worked on the doc only re-use the file used to "patch", that's why it shows up like this

Understood, but then two two should probably be merged together, no?

jburel:
They are merged together, I have not looked at the text on the doc front.
What is coming from omero-install is only ``omero certificates`` the text around is in the doc

dominikl:
Looks good to me 👍

jburel:
Thanks @dominikl

PR #238, opened 2020-04-29
Pg url CentOS 7

jburel:
Use common url to install pg
Check that travis is green

cc @sbesson

### DIFF
--- a/linux/autogenerate.sh
+++ b/linux/autogenerate.sh
@@ -107,6 +107,15 @@ N=$OS
 echo -en '\n' >> $file
 echo "# install Postgres" >> $file
 if [ $OS = "centos7" ] ; then
+    number=$(sed -n '/#start-postgresql-installation-general/=' $dir/step01_"$N"_pg_deps.sh)
+    nrs=$((number+1))
+    number=$(sed -n '/#end-postgresql-installation-general/=' $dir/step01_"$N"_pg_deps.sh)
+    nre=$((number-1))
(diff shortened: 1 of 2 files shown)


jburel:
change in autogen will generate the following output

Screenshot 2020-05-01 at 07 45 41

sbesson (approved):
Changes look good and are inline with https://github.com/ome/ansible-role-postgresql/pull/17/. I am unsure of the drawback of using the versioned RPM versus the latest RPM but we can always revisit.

I use https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-docs-superbuild-push/ and https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-docs-superbuild-build/ to validate the output of this change on the documentation.

Current documentation reads as

Screen Shot 2020-05-01 at 14 16 30

With this PR, the RPM is updated as expected.

Screen Shot 2020-05-01 at 14 16 32

As a side note, I  was under the assumption that these superbuild docs job were running daily but they are not. 

Given the ongoing work on  walkthroughs + OMERO documentation and how much these jobs help with reviewing changes to the walkthroughs, is there a checklist of what remains to be done in order to enable these jobs in the daily Trigger build (and maybe deactive the OMERO-docs job)?

jburel:
From what I remember everything was in place to be included in the daily build, we could try that out with the coming centos8 PR
I will still have to manually copy the guides the first time to the doc PR in order to get travis green then afterwards it will be fine

I think we were thinking to replace the doc job by the one mentioned above

PR #239, opened 2020-05-01
Centos8

jburel:
This PR adds installation instructions for CentOS 8

* Check that travis is green

Note that an URL fix for pg on CenOS 7 is also included this PR, a version bump happened today breaking the build

### DIFF
--- a/linux/step01_centos8_pg_deps.sh
+++ b/linux/step01_centos8_pg_deps.sh
@@ -8,7 +8,7 @@ rm -f /var/lib/rpm/__db*
 
 #start-postgresql-installation-general
 dnf module disable -y postgresql
-yum -y install https://download.postgresql.org/pub/repos/yum/reporpms/EL-8-x86_64/pgdg-redhat-repo-42.0-9.noarch.rpm
+yum -y install https://download.postgresql.org/pub/repos/yum/reporpms/EL-8-x86_64/pgdg-redhat-repo-latest.noarch.rpm
 #end-postgresql-installation-general
 if [ "$PGVER" = "pg96" ]; then
     # Postgres, reconfigure to allow TCP connections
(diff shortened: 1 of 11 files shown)


linux/step01_centos8_pg_deps.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
+
+PGVER=${PGVER:-pg11}
+
+
+#start-postgresql-installation-general
+#required to install postgresql, rpm Db corrupted
+rm -f /var/lib/rpm/__db*
```
joshmoore, line 7: I'm concerned about having this in the final docs. Do we know that this won't match any other (non-PG) directories?

manics: This is the main RPM db. It's possible to rebuild it if it becomes corrupted: https://serverfault.com/questions/94367/centos-rpm-database-trashed-rpm-rebuilddb-wont-fix-can-i-recover-using-v

But if our instructions lead to that corruption (i.e. it's not already corrupted in the base docker/VM image and it's not a yum/dnf bug) I don't think we should publish them.

jburel: I moved it one step down to help during the test so it appears in the doc during that phase
i will revert that commit

jburel: the command ``rm ...`` should no longer appear in the doc now

jburel:
@joshmoore @manics ok to merge so I can fix #237 (certificates config also applies to c8 and I will rename file)? I rolled back the commit helping during review.

jburel:
https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-docs-superbuild-build/68/ available

joshmoore:
WFM

PR #240, opened 2020-05-02
Pg12 support

jburel:
This PR adds support for pg12 on centos7, debian9/10 and ubuntu18.04
This is not marked as the recommended version

Installation is tested via travis
Check that travis is green


dominikl:
Looks good 👍

PR #241, opened 2020-05-12
Python 3.6 ubuntu1604

jburel:
* Add option to install 3.6
The installation is tested via Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ env:
   - ENV=debian10 PGVER=pg12
   - ENV=centos7 PGVER=pg12
   - ENV=ubuntu1604
+  - ENV=ubuntu1604 PYTHONVER=py36
   - ENV=debian9
   - ENV=ubuntu1804
   - ENV=centos7
--- a/linux/autogenerate.sh
+++ b/linux/autogenerate.sh
@@ -195,7 +195,7 @@ echo "#end-step03" >> $file
 
 # create virtual env and install dependencies
 echo -en '\n' >> $file
-echo "#start-step03bis: As omero-server, create a virtual env and install dependencies" >> $file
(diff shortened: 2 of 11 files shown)


dominikl:
Works fine. When run with `PYTHONVER=py36` Python 3.6 is installed additionally, if omitted only 3.5 is installed. However, even if 3.6 is installed the default Python seems to link to 3.5:
```
omero-server@38ccff9906cd:/$ python3 --version
Python 3.5.2
```
Don't know if that's a problem and/or intended.

dominikl:
Sorry, that's the system python, I didn't activate the omero-server virtual env. Testing again...

dominikl:
👍 Using the virtualenv:
```
(venv3) omero-server@557460d0c9fd:/opt/omero/server$ python --version
Python 3.6.10
```

jburel:
@dominikl one could use alternatives to switch python version, I did not want to go that route since we do not know what people want to do

dominikl:
Sure, makes sense!

dominikl:
With `PYTHONVER=py36`:
```
(venv3) root@c4e37797c007:/# python --version
Python 3.6.9
```
👍 
Wonder if that could be a general approach to install it from source, in order to be independant from the system's python version.  Just takes quite a while to compile.

jburel:
installing from source takes a lot of time and many people will avoid it if they can.
That's also why we have built wheels for the Python binding for Ice

jburel:
The last commit cap the version of numpy for ubuntu 16.04+python 3.5 
this could be moved to omero-server if we are happy to cap the version on 16.04 for both 3.5 and 3.6
cc @joshmoore

dominikl:
Looks good to me 👍

PR #245, opened 2020-09-10
debian10: remove conf for openssl

jburel:
Adding support for multiple Ice protocols see https://github.com/ome/omero-py/pull/251
means that modifying the openssl.cnf is no longer required.

After removing the conf changes,
```
su - omero-server -c ". ~/settings.env && omero login -s localhost -p 4064 -u root -w omero_root_password"
WARNING:omero.client:..Ignoring error in client.__del__:<class 'Ice.SocketException'>
InternalException: Failed to connect: Ice.SocketException:
Cannot assign requested address
```

Applying the change from https://github.com/ome/omero-py/pull/251 directly to  ``/opt/omero/server/venv3/lib/python3.7/site-packages/omero/clients.py``

```
su - omero-server -c ". ~/settings.env && omero login -s localhost -p 4064 -u root -w omero_root_password"
Created session for root@localhost:4064. Idle timeout: 10 min. Current group: system
```

cc @mtbc 

Edit
----
 * Travis will not be green until we have a new release of omero with a new version of omero
 * re-activate C8 and ubuntu 20.04 to simplify things since we need to use the daily build to test. Commit 5fa98b8 could be pushed away when we are happy with the state of things.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,8 @@ env:
   - ENV=ubuntu1804
   - ENV=centos7
   - ENV=debian10
-
-jobs:
-  allow_failures:
-    - ENV=ubuntu2004
-    - ENV=centos8
+  - ENV=ubuntu2004
+  - ENV=centos8
 
 
 before_install:  
--- a/linux/step01_debian10_deps.sh
+++ b/linux/step01_debian10_deps.sh
@@ -3,7 +3,3 @@
 apt-get -y install\
     python3 \
     python3-venv
-
-# Fix openssl issues
-sed -e '/MinProtocol/ s/^#*/#/' -i /etc/ssl/openssl.cnf
-sed -e '/CipherString/ s/^#*/#/' -i /etc/ssl/openssl.cnf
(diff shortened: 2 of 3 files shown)


joshmoore:
Generally :+1: for not needing to modify system config to make things work. I could see releasing a patch release off of the previous tag if we need to keep travis green.

jburel:
A patch will be an option to keep travis green but there is still a problem on CentOS 8 
see https://travis-ci.org/github/ome/omero-install/jobs/726206210
ubuntu 20.04 is fine see https://travis-ci.org/github/ome/omero-install/jobs/726206209

jburel:
on C8 docker instance.
tcp connection does not work, using ssl works nicely.

jburel:
Travis is slow today.
See https://travis-ci.org/github/jburel/omero-install/builds/727930527 everything is back to green

jburel:
@mtbc: now that we know that everything works with the daily build. I will push away the commit 1ecd7b7. travis will be read 
but we can merge and be ready for the release since the changes will be needed for the doc.

joshmoore:
As soon as there is an omero-blitz jar, @jburel, I'd be all for getting an RC up on downloads.

PR #248, opened 2020-10-14
change link

jburel:
Modify the link so it can be built using the artifacts from GitHub
The only downside is that the URL will need to be modified every time

### DIFF
--- a/linux/step04_all_omero_install.sh
+++ b/linux/step04_all_omero_install.sh
@@ -19,7 +19,7 @@ if [ "$ICEVER" = "ice36" ]; then
 	if [ $OMEROVER == "latest" ]; then
 		#start-release-ice36
 		cd /opt/omero/server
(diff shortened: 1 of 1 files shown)


manics:
https://github.com/ome/omego/pull/127 could help. It currently requires the full tag, but in future this could be changed to list all tags and do semver-based matching of partial versions.

jburel:
@manics I used omego previously for merge-build etc. Happy to bring that back in

manics:
We could get this PR in now assuming everything works, then switch to omego separately

jburel:
Travis is green so it works!!

manics (approved):

jburel:
Another option will be to drop the version from the zip attached on GitHub and use

https://github.com/ome/openmicroscopy/releases/latest/download/OMERO.server.zip

PR #244, opened 2020-09-08
use individual packages instead of meta-package

jburel:
use individual packages instead of meta-package
cc @sbesson 

Check that travis is green

PR #249, opened 2020-11-11
Gh actions

jburel:
Replace travis by GH actions
See https://github.com/jburel/omero-install/actions/runs/358365632

Removing the ``it`` flag was necessary to avoid ``no tty`` error when using GH actions


sbesson (approved):
Changes and the the ouptut of the workflow looks good.

The `-it` argument had been originally introduced as part of https://github.com/ome/omero-install/pull/244#pullrequestreview-484734896 to allow the commands to be displayed in Travis CI logs but it looks like they are unnecessary in the GitHub environment :+1:

Only question is whether we want to keep this as a matrix workflow vs n workflows. Advantage of the former is definitely simplicity and maintenance. THe main reason for considering the latter is that I don't think re-running a single matrix build is possible at the moment. Maybe we get this in its current form and revisit it if builds are proving to be regularly flaky?

jburel:
I have opted for the matrix approach since it simplifies things. I have followed the same approach for omeroweb-install.
I think it makes the files easier to read and maintain.
We can always revisit later on

PR #252, opened 2021-04-27
omero-certificates

jburel:
Configure ``omero-certifactes`` on CentOS 7, 
Without the changes, it leads to https://github.com/ome/omero-install/pull/251/checks?check_run_id=2446211596

cc @sbesson

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,7 +1,11 @@
 ---
 name: Build
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 0 * * 0'
 
 jobs:
   build:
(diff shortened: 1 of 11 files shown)


joshmoore:
Might be worth renaming the file to "install-omero-certicates" since we're no longer patching.

jburel:
good point. Note that the change will modify a good number of files. I will push a commit

sbesson:
Or include it as part of `step04_all_omero.sh` ? THis would mirror the behavior in https://github.com/ome/ansible-role-omero-server/blob/master/templates/00-omero-server-omero.j2

jburel:
I will check the impact on the doc since renaming might be the easiest at this stage

jburel:
Any objection to merge this PR? remaining failure in https://github.com/ome/omero-install/pull/255 will be fixed when this PR is merged

joshmoore:
WFM.

sbesson (approved):
The installation of `openssl` and the addition of `certificates` aligns closely with the behavior of the Ansible role - see https://github.com/ome/ansible-role-omero-server/pull/50 and https://github.com/ome/ansible-role-omero-server/pull/58.

As a bonus this removes some of the CentOS specific logic in the auto-generation script and reduces the complexity.

PR #255, opened 2021-12-07
Remove ``omero-py`` install step. Already pulled via ``omero-server[x]``

jburel:
* Remove ``omero-py`` installation step. This is already installed by ``pip install omero-server[x]``
* Upgrade ``pip`` otherwise the installation of the python dependencies using ``pip`` fail
* Upgrade centos7 image. Change to be applied to https://github.com/ome/omero-ssh-c7-docker

### DIFF
--- a/linux/step04_all_omero_install.sh
+++ b/linux/step04_all_omero_install.sh
@@ -6,13 +6,6 @@ set -eux
 
 . `dirname $0`/settings.env
 
-#start-install-omero-py
-# Install omero-py
-$VENV_SERVER/bin/pip install "omero-py>=5.8.0"
-#end-install-omero-py
-
-#start-download-omero
-
 
 #start-install
 if [ "$ICEVER" = "ice36" ]; then
--- a/linux/autogenerate.sh
+++ b/linux/autogenerate.sh
@@ -176,13 +176,7 @@ echo "$line" >> $file
 echo "#end-step03bis" >> $file
 
 echo -en '\n' >> $file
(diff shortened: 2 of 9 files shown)


sbesson:
The upgrade of `pip` to comply with recent package versions matches what has been introduced in https://github.com/ome/ansible-role-omero-server/pull/61. :+1:

Note a similar upgrade will likely be required in omeroweb-install - see https://github.com/ome/ansible-role-omero-web/pull/41

linux/step04_all_omero_install.sh
```diff
@@ -6,13 +6,6 @@ set -eux
 
 . `dirname $0`/settings.env
 
-#start-install-omero-py
-# Install omero-py
-$VENV_SERVER/bin/pip install "omero-py>=5.8.0"
```
sbesson, line 11: I assume this cleanup is motivated by the fact `omero-py` is pulled as a server dependency via `pip install omero-server[default]` (the title of the PR is slightly frightening)?

jburel: it is.
The title got picked up from commit message (partial). Updating

jburel:
> Note a similar upgrade will likely be required in omeroweb-install - see [ome/ansible-role-omero-web#41](https://github.com/ome/ansible-role-omero-web/pull/41)

see https://github.com/ome/omeroweb-install/pull/77

sbesson (approved):

Issue #256, opened 2021-12-17
Python 3.5 end of Life Debian 9


jburel:
Python 3.5 is the default python version installed on debian 9 i.e. ``apt-get -y install python3``. Python 3.5 has reached end of life (13/09/2021)
To install Python 3.6, this will have to be done from source. Making the installation instructions more complicated than needed.
Security support has stopped for Debian 9 see https://endoflife.date/debian

Installation of OMERO on Debian 9 should be removed from the list of operating systems See https://docs.openmicroscopy.org/omero/5.6.3/sysadmins/unix/server-installation.html
cc @sbesson @joshmoore 

Repositories to update:

- [x] https://github.com/ome/omero-install
- [x] https://github.com/ome/omeroweb-install
- [x] https://github.com/ome/omero-documentation

Edit: Support for Debian11 should be considered. This is outside the scope of this issue

sbesson:
Makes sense to me. Debian is not directly listed in https://github.com/ome/omero-documentation/blob/develop/omero/sysadmins/version-requirements.rst so I suspect our level of support is lower than RHEL/Ubuntu. Also, looking briefly, I think this EOL has no effect the level of support for other distributions since we already dropped support for Ubuntu 16.

Probably even more impactful is the upcoming EOL of Python 3.6 (https://endoflife.date/python). We already saw some of the effect with many of the packages now supported the last three minor releases. I assume this will need to get captured and force a wider discussion around supported platforms.

jburel:
> Probably even more impactful is the upcoming EOL of Python 3.6 (https://endoflife.date/python). We already saw some of the effect with many of the packages now supported the last three minor releases. I assume this will need to get captured and force a wider discussion around supported platforms.

Yes this is a problem, excepting Debian 10, 3.6 is installed by default when installing Python 3 in all OS.

PR #257, opened 2021-12-17
Remove Debian9

jburel:
See https://github.com/ome/omero-install/issues/256


sbesson (approved):
The main question from my side is the impact of this removal on published documentation pages - see https://github.com/ome/omero-documentation/pull/2207#issuecomment-1005720645

jburel:
I will take care of the documentation and point to the Debian 10 installation

PR #258, opened 2022-01-05
CentOS 8 and Python 3.8

jburel:
Changes required to run OMERO on CentOS 8 with Python 3.8 (default is 3.6)
The wheel/zip currently available at https://github.com/jburel/zeroc-ice-centos8/releases/tag/0.0.2 will need to be moved to ``ome``.

cc @joshmoore @sbesson

### DIFF
--- a/linux/step01_centos8_deps.sh
+++ b/linux/step01_centos8_deps.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-yum -y install python3
+yum -y install python38
 yum -y install openssl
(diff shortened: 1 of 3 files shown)


sbesson:
Works for me, it looks like https://github.com/ome/omero-install/pull/257 will need to be integrated in order to get GHA CI passing here as well.
Should we open your wheel changes as a PR to link the two of them together?

sbesson:
See https://github.com/ome/omero-install/issues/262. The fate of CentOS 8 support is still TBD

PR #259, opened 2022-01-10
Possible option: Conda

jburel:
This PR installa omero-py in a conda environment. This allows us to use newer version of Python.
This has been tested with Python 3.9
cc @joshmoore @sbesson

PR #260, opened 2022-01-10
Possible option: Scl python 3.8 on CentOS 7

jburel:
This PR installs scl python 3.8 
No wheel used in that option

cc @joshmoore @sbesson

### DIFF
--- a/linux/settings.env
+++ b/linux/settings.env
@@ -17,4 +17,4 @@ export OMERODIR=/opt/omero/server/OMERO.server
 # Location of the virtual environment for omero-py
 VENV_SERVER=/opt/omero/server/venv3
 
-export PATH=$VENV_SERVER/bin:/opt/rh/rh-python38/root/usr/bin:$PATH
+export PATH=$VENV_SERVER/bin:$PATH
--- a/linux/settings.env
+++ b/linux/settings.env
@@ -17,4 +17,4 @@ export OMERODIR=/opt/omero/server/OMERO.server
 # Location of the virtual environment for omero-py
 VENV_SERVER=/opt/omero/server/venv3
 
-export PATH=$VENV_SERVER/bin:$PATH
+export PATH=$VENV_SERVER/bin:/opt/rh/rh-python38/root/usr/bin:$PATH
--- a/linux/step01_centos7_ice_venv.sh
+++ b/linux/step01_centos7_ice_venv.sh
@@ -11,7 +11,7 @@ python3 -mvenv $VENV_SERVER
 $VENV_SERVER/bin/pip install --upgrade pip
 
 # Install the Ice Python binding
-$VENV_SERVER/bin/pip install https://github.com/jburel/zeroc-ice-py-centos7/releases/download/0.3.0/zeroc_ice-3.6.5-cp38-cp38-linux_x86_64.whl
+$VENV_SERVER/bin/pip install https://github.com/ome/zeroc-ice-py-centos7/releases/download/0.3.0/zeroc_ice-3.6.5-cp38-cp38-linux_x86_64.whl
 
 # Install server dependencies
 $VENV_SERVER/bin/pip install omero-server[default]
--- a/linux/step01_centos7_ice_venv.sh
+++ b/linux/step01_centos7_ice_venv.sh
@@ -5,7 +5,7 @@ set -e -u -x
 VENV_SERVER=${VENV_SERVER:-/opt/omero/server/venv3}
 #start-ice-py
 # Create a virtual env
-python -mvenv $VENV_SERVER
+python3 -mvenv $VENV_SERVER
 
 # Upgrade pip
 $VENV_SERVER/bin/pip install --upgrade pip
--- a/linux/step01_centos7_ice_venv.sh
+++ b/linux/step01_centos7_ice_venv.sh
@@ -11,7 +11,7 @@ python3 -mvenv $VENV_SERVER
 $VENV_SERVER/bin/pip install --upgrade pip
 
 # Install the Ice Python binding
-#$VENV_SERVER/bin/pip install https://github.com/ome/zeroc-ice-py-centos7/releases/download/0.2.1/zeroc_ice-3.6.5-cp36-cp36m-linux_x86_64.whl
+$VENV_SERVER/bin/pip install https://github.com/jburel/zeroc-ice-py-centos7/releases/download/0.3.0/zeroc_ice-3.6.5-cp38-cp38-linux_x86_64.whl
 
 # Install server dependencies
 $VENV_SERVER/bin/pip install omero-server[default]
--- a/linux/step01_centos7_ice_venv.sh
+++ b/linux/step01_centos7_ice_venv.sh
@@ -5,7 +5,7 @@ set -e -u -x
 VENV_SERVER=${VENV_SERVER:-/opt/omero/server/venv3}
 #start-ice-py
 # Create a virtual env
-python3 -mvenv $VENV_SERVER
+python3 -mvenv --upgrade $VENV_SERVER
 
 # Upgrade pip
 $VENV_SERVER/bin/pip install --upgrade pip
--- a/linux/step01_centos7_ice_venv.sh
+++ b/linux/step01_centos7_ice_venv.sh
@@ -5,7 +5,7 @@ set -e -u -x
 VENV_SERVER=${VENV_SERVER:-/opt/omero/server/venv3}
 #start-ice-py
 # Create a virtual env
-python3 -mvenv --upgrade $VENV_SERVER
+/opt/rh/rh-python38/root/usr/bin/python -mvenv --upgrade $VENV_SERVER
 
 # Upgrade pip
 $VENV_SERVER/bin/pip install --upgrade pip
--- a/linux/step01_centos7_ice_venv.sh
+++ b/linux/step01_centos7_ice_venv.sh
@@ -5,13 +5,13 @@ set -e -u -x
 VENV_SERVER=${VENV_SERVER:-/opt/omero/server/venv3}
 #start-ice-py
 # Create a virtual env
-python3 -mvenv $VENV_SERVER
+python -mvenv $VENV_SERVER
 
 # Upgrade pip
 $VENV_SERVER/bin/pip install --upgrade pip
 
 # Install the Ice Python binding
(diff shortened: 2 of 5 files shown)


linux/step01_centos7_ice_venv.sh
```diff
@@ -5,13 +5,13 @@ set -e -u -x
 VENV_SERVER=${VENV_SERVER:-/opt/omero/server/venv3}
 #start-ice-py
 # Create a virtual env
-python3 -mvenv $VENV_SERVER
+python -mvenv $VENV_SERVER
```
sbesson: Arg, there is no `python3` symlink? That's really unfortunate in terms of consistency

linux/step01_centos7_ice_venv.sh
```diff
@@ -5,13 +5,13 @@ set -e -u -x
...
 
 # Upgrade pip
 $VENV_SERVER/bin/pip install --upgrade pip
 
 # Install the Ice Python binding
-$VENV_SERVER/bin/pip install https://github.com/ome/zeroc-ice-py-centos7/releases/download/0.2.1/zeroc_ice-3.6.5-cp36-cp36m-linux_x86_64.whl
+#$VENV_SERVER/bin/pip install https://github.com/ome/zeroc-ice-py-centos7/releases/download/0.2.1/zeroc_ice-3.6.5-cp36-cp36m-linux_x86_64.whl
```
sbesson: Is the issues  that we have not built a wheel yet for testing or is a wheel not an option in this scenario?

linux/step01_centos7_ice_venv.sh
jburel: There is a symlink I think. I was testing various thing

linux/step01_centos7_ice_venv.sh
jburel: we have to build a wheel which is something we may want to reconsider.
This obviously has an impact on the dependencies required e.g. openssl-devel, python-devel etc.

jburel: c2cacd9 uses the wheel

jburel:
general consensus is to go with that option.

sbesson:
Few comments from today's discussion:
- comparing this workflow with #259, immediate impressions is that the SCL solution is the less disruptive
- it might be worth having a reference issue where we can have a table of pros/cons before making the decision
- strong argument towards deciding on a single solution for our docs/roles rather than attempting to support multiple options
- we can certainly consider recommending e.g. SCL for the sysadmin workflows vs Conda for the developer workflows
- @jburel will look into OMERO.web next
- the preferred solution can be investigated via CI as a next step
- some Ansible role upgrades can be pre-tested on internal servers e.g. `ome-training-3` and potentially new IDR deployments
- for existing deployments, the problem of the virtual environment upgrade (from Python 3.6 to Python 3.8) will need to be taken into account

joshmoore:
> * for existing deployments, the problem of the virtual environment upgrade (from Python 3.6 to Python 3.8) will need to be taken into account

Perhaps we start naming the environments explicitly? e.g. `/opt/...python38-scl`

sbesson:
> Perhaps we start naming the environments explicitly? e.g. /opt/...python38-scl

Agreed, we also discussed options briefly but this is a point of investigation. The main downside of creating a new environment is the potential confusion in existing deployments e.g. we already have deployments with both `/opt/omero/server/venv` (Python 2) and `/opt/omero/server/venv3` (Python 3). On the other side, this probably keeps both the documentation and the deployment infrastructure as simple as possible leaving former virtual environments stale.

jburel:
I think having a clean env will be the most suitable option. Packages in venv3, will unlikely be compatible with the new python version. This could lead to a big mess. Renaming to ``python38-scl`` seems to be safest option.

jburel:
Changing the name implies introducing a new ``settings.env`` file and make the infra a bit more complex.

sbesson:
From the Ansible perspective, in the absence of built-in method to invalidate an existing virtual environment with a different Python version, a new environment feels like the simplest approach to support both new and existing deployments and meet the idempotence requirement.

One consideration though: will the virtual environment name become OS specific? This would have the downside of creating divergence in files like https://github.com/ome/omero-install/blob/a2d7b36cddd8995355a96f3fa6f7472f4cb72392/linux/omero-server-systemd.service#L20

jburel:
> One consideration though: will the virtual environment name become OS specific? This would have the downside of creating divergence in files like
> 
> https://github.com/ome/omero-install/blob/a2d7b36cddd8995355a96f3fa6f7472f4cb72392/linux/omero-server-systemd.service#L20

That one too

linux/settings.env
```diff
@@ -17,4 +17,4 @@ export OMERODIR=/opt/omero/server/OMERO.server
 # Location of the virtual environment for omero-py
 VENV_SERVER=/opt/omero/server/venv3
 
-export PATH=$VENV_SERVER/bin:$PATH
+export PATH=$VENV_SERVER/bin:/opt/rh/rh-python38/root/usr/bin:$PATH
```
sbesson: One consideration while looking again at this.

Since the entire installation/deployment process is encapsulated using Python virtual environment, it is possible that updating the environment `PATH` is not strictly necessary. At minimum, the `python3 -mvenv` command in `step01_centos7_ice_venv.sh` needs to know about the Python 3.8 installation but I do not think this is a required later in the deployment process.

The reason for mentioning this is that I was thinking of equivalent changes at the Ansible level and I could imagine introducing some RH7 specific arguments in https://github.com/ome/ansible-role-python3-virtualenv/blob/947d49c596fa5fee3318d8c8d7d5b4b4bd3f9899/files/ome-python3-virtualenv.sh#L19

linux/test/centos7/Dockerfile
```diff
@@ -29,4 +29,5 @@ ARG PGVER=pg11
 RUN cd omero-install-test && \
 	bash install_centos7.sh
 
+
```
sbesson: Probably an oversight?

sbesson (commented):
One (untested) comment about the injection of the Python 3.8 path.

Although I think that `omero-install` only covers _new_ installations, the biggest remaining question is the naming of the virtual environment and the handling of existing Python 3.6 virtual environments.

Now that the former work has been integrated and tagged, we might want to start integrating this work into the development branch of `omero-install`.

linux/settings.env
jburel: I oped to add i to ``settings.env`` since our doc we indicate to run ``source settings.env`` so making the change will not require an export PATH before running the command

sbesson: Another option if we don't want to export the PATH would be to explicitly set the path to the executable in the CentOS 7 file e.g.

```
/opt/rh/rh-python38/root/usr/bin/python -m venv ...
```

In addition to the point above, my other concern is whether this will appear in Ubuntu walkthroughs as well as it might create confusion

jburel: that will also change the doc. I understood we wanted to keep ``python3 -m venv..``

jburel:
According to the [documentation](https://docs.python.org/3/library/venv.html), using ``--upgrade`` should upgrade the virtual env
i.e. ``python3 -m venv --upgrade $VENV_SERVER``

jburel:
Trying to upgrade an existing env will lead to problem  due for example to package like
```
$VENV_SERVER/bin/pip install https://github.com/ome/zeroc-ice-py-centos7/releases/download/0.2.1/zeroc_ice-3.6.5-cp36-cp36m-linux_x86_64.whl
```
If other dependencies have been installed from a wheel using a specific Python version, it will be the same problem
We will need to add information at the doc level.

sbesson:
Thanks. That's definitely the downside of wheels which are tied to a specific Python version. Off-hand, I do not see a mechanism by which an upgrade flag could be aware of the new wheel to install.

I suspect if we go down the SCL route, we will have to go for a fresh environment either way. Coming back to the path issue, I am leaning towards  not changing the virtual environment path. 

One thought: is there a command that would allow to programmatically tell whether a  virtual environment was built using a given Python version matching an executable path? If so, I could certainly imagine a Python upgrade worfklow similar to:

1. install SCLPython
2. run the Python version compatibility command
3. if incompatible
  a. archive the former environment (might be as simple as `mv venv3 venv3-old`).
  b. optionally try to create a report of the requirements via `pip freeze`?
  c. create a  new virtual environment under `venv3`
  d. depending on b), try to reinstall the previous requirements (this might still fail for wheels)
4. install the requirements and start the server

For the Ansible upgrade workflow, I could see implementing 2, 3a and 3c as part of the ome.python3_virtualenv role. The requirements freeze/reinstallation (3b, 3d) should be unnecessary in this case as the Python dependencies should be fully captured using roles & variables and a re-execution of the playbook should re-install everything under the new environment.

sbesson:
I have performed additional investigation on the Ansible workflow. The current logic for the Python environment management is the following:

- `ome.python3_virtualenv` installs the  base Python as well as a shared wrapper script allowing to create a virtual envrionment
- `ome.omero_server` and `ome.omero_web` are consuming this wrapper script via the `virtualenv_command` parameter of the [pip](https://docs.ansible.com/ansible/2.8/modules/pip_module.html) module
- the `virtualenv_command`  is only invoked if the virtual environment does not exist. There is no built-in Ansible parameter that solves the issue of an upgrade where a virtual environment exist but must be recreated/upgrade

I also looked at some options of the `venv` command (https://docs.python.org/3.6/library/venv.html):

- `--upgrade` has requirements on the Python upgrade which do not apply to the SCL case and would likely not handle the upgrade of wheels as per the discussion above
- `--clear` is largely the equivalent of `rm -rf VENV_DIR && python -mvenv VENV_DIR`. It will create a fresh upgrade environment but does not solve the problem of the one-off upgrade

Based on the above, it feels like there is not built-in native solution to achieve what we want. I would also be very conservative about building new tools.
A possible approach would be to treat this type of upgrade similarly to e.g. a PostgreSQL version upgrade in the sense that the process requires more than just updating requirements and remotely executing the playbooks. Minimally, the following worfklow should be sufficient:

- stop the server/web services via `service stop`
- delete/rename the existing `/opt/ome/{server.,web}/venv3` environments
- update the roles to install SCL Python and the associated wheels
- run the Ansible playbooks

jburel:
Closing this PR
A branch was created so it can be used in the current devspace

PR #261, opened 2022-02-01
Check for omero-server user

jburel:
Check for the ``omero-server`` user not the ``omero`` user.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        os: [centos7, centos8, debian10, ubuntu1804, ubuntu2004]
+        os: [centos7, debian10, ubuntu1804, ubuntu2004]
         pg: [pg11, pg12]
     steps:
       - name: Checkout
--- a/linux/install_centos7.sh
+++ b/linux/install_centos7.sh
@@ -27,15 +27,21 @@ if [[ "$PGVER" =~ ^(pg96|pg10|pg11|pg12)$ ]]; then
     bash -eux step03_all_postgres.sh
 fi
 
-cp settings.env step04_all_omero.sh setup_omero_db.sh ~omero-server
-
 bash -eux step01_centos7_ice_venv.sh
+
+if [ "$(getent passwd omero-server)" ]; then
+    cp settings.env step04_all_omero.sh setup_omero_db.sh ~omero-server
+fi
+
+
 
 OMEROVER=$OMEROVER ICEVER=$ICEVER bash -eux step04_all_omero_install.sh
 
-su - omero-server -c " bash -eux step04_all_omero.sh"
+if [ "$(getent passwd omero-server)" ]; then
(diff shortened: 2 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        os: [centos7, centos8, debian10, ubuntu1804, ubuntu2004]
+        os: [centos7, debian10, ubuntu1804, ubuntu2004]
```
sbesson, line 15: Is that the recommendation for https://github.com/ome/omero-install/issues/262?

linux/install_centos7.sh
```diff
@@ -27,15 +27,21 @@ if [[ "$PGVER" =~ ^(pg96|pg10|pg11|pg12)$ ]]; then
     bash -eux step03_all_postgres.sh
 fi
 
-cp settings.env step04_all_omero.sh setup_omero_db.sh ~omero-server
-
 bash -eux step01_centos7_ice_venv.sh
 
+if [ "$(getent passwd omero-server)" ]; then
+    cp settings.env step04_all_omero.sh setup_omero_db.sh ~omero-server
```
sbesson, line 36: Can block be moved with the similar `if` block below or do some steps for `step04_all_omero_install.sh` depend on these files?

linux/install_centos7.sh
```diff
@@ -27,15 +27,21 @@ if [[ "$PGVER" =~ ^(pg96|pg10|pg11|pg12)$ ]]; then
...
+fi
+
+
+
 OMEROVER=$OMEROVER ICEVER=$ICEVER bash -eux step04_all_omero_install.sh
 
-su - omero-server -c " bash -eux step04_all_omero.sh"
+if [ "$(getent passwd omero-server)" ]; then
```
sbesson: Am I correct that the goal of this extra logic is to have `install_centos7.sh` as the single entry-point but with support different installation paths?

Is it possible to exaplin why these steps are skipped?

linux/step02_all_setup.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-if [ -z "$(getent passwd omero)" ]; then
+if [ -z "$(getent passwd omero-server)" ]; then
```
sbesson, line 3: Is an `omero` user still created anywhere? Or is that cleaning legacy check?

.github/workflows/main.yml
jburel: If we want to support CentOS 8, this will require extra work outside the scope of this PR
This change does not build.
I am for removing it

linux/step02_all_setup.sh
jburel: an ``omero`` user is created in the parent Docker. 
This is an mistake when we split the omero user into ``omero-server`` and ``omero-web``

sbesson (approved):

jburel:
Tagging as 5.6.4 and bumping version in https://github.com/ome/devslave-c7-docker so I can remove the usage of personal repository

PR #264, opened 2022-03-23
Pg96

jburel:
* Remove ref to pg 9.6 
* fix default in ubuntu20.04

Nothing to test. This is tested via GHA

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,6 +14,9 @@ jobs:
       matrix:
         os: [centos7, debian10, ubuntu1804, ubuntu2004]
         pg: [pg11, pg12]
+        exclude:
+          - os: ubuntu2004
+            pg: pg11
     steps:
       - name: Checkout
         uses: actions/checkout@v1
@@ -24,7 +27,6 @@ jobs:
           cd linux
           ./autogenerate.sh
       - name: Build docker image
-        if: contains(matrix.pg, 'pg12') || (!contains(matrix.os, 'ubuntu2004') && contains(matrix.pg, 'pg11'))
         run: |
           cd linux/test
           PGVER=$PGVER ./docker-build.sh $ENV
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,6 +24,7 @@ jobs:
           cd linux
           ./autogenerate.sh
       - name: Build docker image
+        if: contains(matrix.pg, 'pg12') || (!contains(matrix.os, 'ubuntu2004') && contains(matrix.pg, 'pg11'))
         run: |
           cd linux/test
           PGVER=$PGVER ./docker-build.sh $ENV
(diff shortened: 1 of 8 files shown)


linux/install_debian10.sh
```diff
@@ -27,7 +27,7 @@ bash -eux step01_debian10_pg_deps.sh
 
 bash -eux step02_all_setup.sh
 
-if [[ "$PGVER" =~ ^(pg96|pg10|pg11|pg12)$ ]]; then
+if [[ "$PGVER" =~ ^(pg11|pg12)$ ]]; then
```
sbesson, line 30: Any reason why `pg10` is removed here?

linux/install_ubuntu2004.sh
```diff
@@ -3,7 +3,7 @@
 set -e -u -x
 
 OMEROVER=${OMEROVER:-latest}
-PGVER=${PGVER:-pg11}
+PGVER=${PGVER:-pg12}
```
sbesson, line 6: Why the switch to `pg12` while Debian / CentOS 7 are still at `pg11` by default?

sbesson (commented):
The removal of the PostgreSQL 9.6 part makes sense. The bulk of my remaining questions is around the update of the default/supported values for the Debian/Ubuntu platforms

linux/install_ubuntu2004.sh
jburel: pg12 is the default version installed. This is a bug
we do not have instructions on how to install pg11

linux/install_debian10.sh
jburel: only pg11 and pg12 recipes available

linux/install_ubuntu2004.sh
sbesson: Indeed I just realized that https://github.com/ome/omero-install/runs/5110966381 is effectively installing PostgreSQL 12 under the hood.

Since the default is updated in this PR, can we update the GitHub workflow to skip this combination and match the fact only the default version is supported/tested?

jburel: [711ef74](https://github.com/ome/omero-install/pull/264/commits/711ef74fedd1e939841e87bb911d13e92279c605) will not install with pg11 on ubuntu2004

.github/workflows/main.yml
```diff
@@ -24,6 +24,7 @@ jobs:
           cd linux
           ./autogenerate.sh
       - name: Build docker image
+        if: contains(matrix.pg, 'pg12') || (!contains(matrix.os, 'ubuntu2004') && contains(matrix.pg, 'pg11'))
```
sbesson: It took me a while to figure out the check here. Do I read that correctly as PSQL 12 for any platform + PSQL 11 for any platform except Ubuntu 20.04?

Possibly this logic could be moved alongside the matrix metadata above using the [exclude](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-configurations-from-a-matrix) keyword to ease maintenance?

jburel: You read it correctly. Exclude will make things cleaner.

sbesson (approved):

PR #265, opened 2022-04-26
reset generic url

jburel:
Use downloads instead of GH


jburel:
Green, merging before tagging

PR #266, opened 2022-07-04
use omego

jburel:
To avoid update of the documentation when new version will be release e.g. 5.7.0
the command will install the latest version in the 5 series

### DIFF
--- a/linux/step01_centos7_ice_venv.sh
+++ b/linux/step01_centos7_ice_venv.sh
@@ -10,6 +10,9 @@ python3 -mvenv $VENV_SERVER
 # Upgrade pip
 $VENV_SERVER/bin/pip install --upgrade pip
 
+# Install omego to install the server
+$VENV_SERVER/bin/pip install --upgrade omego
+
 # Install the Ice Python binding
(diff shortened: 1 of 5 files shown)


dominikl:
Looks good to me. Tested it only with Debian10, but as it's all the same for the other distros, that should be enough, isn't it?

jburel:
As discussed in https://github.com/ome/openmicroscopy/issues/6330, the main question is if we want to expose ``omego`` or not in the installation steps

@dominikl the installation is tested via GHA

joshmoore:
For that diff, @jburel, it might not be warranted. Are there other locations you have in mind for using `omego`?

jburel:
Not really, But if we have https://github.com/ome/openmicroscopy/pull/6318 in without any build number
I could easily:
* find the latest version from GitHub
*  build the URL from GitHub.

i.e. no usage of ``omego``
I was planning initially to use it https://github.com/ome/omero-documentation/pull/2261 but I am happy with the alternative solution.

jburel:
Closing this PR. I don't think I will be using ``omego``. With https://github.com/ome/openmicroscopy/pull/6318 in, build number will not be present so it will be possible to build a URL.

PR #273, opened 2023-07-18
Add rocky linux 9

dominikl:
See title. 
Uses @sbesson 's Ice builds: https://github.com/sbesson/zeroc-ice-rockylinux9-x86_64/tree/main and https://github.com/sbesson/zeroc-ice-py-rockylinux9-x86_64


jburel:
replaced by https://github.com/ome/omero-install/pull/274

PR #274, opened 2023-08-29
Rocky linux 9

jburel:
This PR replaces https://github.com/ome/omero-install/pull/273

Make changes to work in docker
Tested via GHA
cc @dominikl

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,14 +12,14 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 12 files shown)


dominikl:
Looks good. Installation via [install_rocky9.sh](https://github.com/jburel/omero-install/blob/rocky/linux/install_rocky9.sh) into a Rocky9 VM still works fine too 👍

jburel:
Merging so we can start working on doc

jburel:
url to Ice will need to be updated

PR #277, opened 2023-09-25
Simplify rocky9 installation

dominikl:
Some changes to the rocky9 installation scripts:
- Moved the commands from ...java_deps.sh, ...ice_dep.sh and ...pg_deps.sh into step01_rocky9_deps.sh. 
- Removed the dependencies on centos7 scripts.
- Use `dnf` instead of `yum` throughout.

Includes the change from https://github.com/ome/omero-install/pull/276 .


linux/step01_rocky9_ice_venv.sh
```diff
@@ -5,7 +5,7 @@ set -e -u -x
 VENV_SERVER=${VENV_SERVER:-/opt/omero/server/venv3}
 #start-ice-py
 # Create a virtual env
-python3 -mvenv $VENV_SERVER
+python -mvenv $VENV_SERVER
```
jburel: This will not work on rocky9
I had to fix several of them in devspace

dominikl:
I give up now, trying to get this PR through the docker workflow. I think we should just scrap these docker tests. What is the point in the test if you have to constantly diverge in the script "if docker do this, if not do that", just to get the workflow to pass? It works on a real linux system.

dominikl:
Just tested with Petr against the rhel9 VM. The blocked ports are not a problem of the VM, but of rhel9 itself. 
Have to run `firewall-cmd --zone=public --add-port=4064/tcp --permanent` so that omero-server is accessible on the ip directly. Should that be added to the scripts as well, as part of the default installation @jburel ?
Resp for web `firewall-cmd --zone=public --add-port=4080/tcp --permanent` .

(followed by `firewall-cmd --reload`)

linux/install_rocky9.sh
jburel, line 1: We will need to move that so we can include it in the documentation

jburel: done

dominikl:
👍 Thanks

PR #278, opened 2023-11-08
fix generation of doc for rocky9

jburel:
Remove duplicate entry in the doc
We currently have
```
yum -y install epel-release

yum -y install unzip wget bc

# install Java
dnf -y install java-11-openjdk

# install dependencies
dnf -y install python unzip bzip2 wget bc openssl
```

Now we should have 

```
# install Java
dnf -y install java-11-openjdk

# install dependencies
dnf -y install python unzip bzip2 wget bc openssl
```


dominikl (approved):
👍 Thanks. Makes sense!

PR #279, opened 2023-11-15
recommend psql 14

jburel:
Change the recommended version following yesterday's discussion


dominikl (approved):
👍

pwalczysko:
@jburel on both training as well as learning playbook based on rhel9 roles I have postgres 14, and the ugm server is indeed on 14 (checked from within the server), so it should be matching

PR #284, opened 2023-11-29
use jre instead of jdk

jburel:
Notice while checking https://github.com/ome/omero-documentation/pull/2383


dominikl (approved):
True, JDK isn't actually required 👍

PR #289, opened 2024-07-05
Debian10

jburel:
Remove installation for debian 10 (EOL 30/06/2024)


dominikl (approved):
Makes sense, thanks @jburel 
Are we planning to add latest debian again or only support rocky on the long run?

jburel:
No Plan to add the latest debian for now

PR #292, opened 2024-11-04
Change location where to download the file from

jburel:
During a release the artifacts from Github and copy them to Dundee
Installing the server from non EU countries make the installation process painfully slow. This was noticed while installing from Japan
This PR installs the server using directly the zip from gh
Note that in order to avoid an update every time a new version of OMERO.server is released, the version number needs to be dropped, see https://github.com/ome/openmicroscopy/pull/6414

### DIFF
--- a/linux/step04_all_omero_install.sh
+++ b/linux/step04_all_omero_install.sh
@@ -8,9 +8,11 @@ set -eux
 
 #start-install
 if [ $OMEROVER == "latest" ]; then
+	# version number and ice version to be dropped
 	#start-release-ice36
 	cd /opt/omero/server
-	SERVER=https://downloads.openmicroscopy.org/omero/5.6/server-ice36.zip
+	SERVER=https://github.com/ome/openmicroscopy/releases/latest/download/OMERO.server-5.6.13-ice36.zip
+
 	wget -q $SERVER -O OMERO.server-ice36.zip
 	unzip -q OMERO.server*
 	#end-release-ice36


linux/step04_all_omero_install.sh
```diff
@@ -8,9 +8,11 @@ set -eux
 
 #start-install
 if [ $OMEROVER == "latest" ]; then
+	# version number and ice version to be dropped
 	#start-release-ice36
 	cd /opt/omero/server
-	SERVER=https://downloads.openmicroscopy.org/omero/5.6/server-ice36.zip
+	SERVER=https://github.com/ome/openmicroscopy/releases/latest/download/OMERO.server-5.6.13-ice36.zip
```
pwalczysko, line 14: ``OMERO.server-5.6.13-ice36.zip`` seems hardcoded ? Maybe I misunderstand, what will happen once we have OMERO.server-5.6.14 ?

linux/step04_all_omero_install.sh
```diff
@@ -8,9 +8,11 @@ set -eux
 
 #start-install
 if [ $OMEROVER == "latest" ]; then
+	# version number and ice version to be dropped
```
pwalczysko, line 11: yes, who and when will drop them ?

linux/step04_all_omero_install.sh
jburel: The PR https://github.com/ome/openmicroscopy/pull/6414 will drop the version 
and push OMERO.server.zip
This means that we won't need to upgrade it every time

It is currently hard-coded otherwise the build will not work

linux/step04_all_omero_install.sh
jburel: Next release if https://github.com/ome/openmicroscopy/pull/6414 is accepted and included

jburel:
Closing 
This will be handled via nginx redirect

PR #293, opened 2024-11-04
Review Python installation

jburel:
This should fix the build failure noticed with the latest jobs on rocky9


jburel:
Merging so we can fix build in other PRs

PR #296, opened 2025-03-05
Do not test on pg13

jburel:
Remove pg13 
This is failing the build

Check that the build is green


jburel:
merging so the build can be green again

dominikl:
Shall we remove PG 13 (we even list 12 in there) from https://omero.readthedocs.io/en/stable/sysadmins/version-requirements.html#omero-support-policies too?

PR #299, opened 2025-12-01
Update rocky9 install

dominikl:
See title. Mostly install python 3.12 and bump postgres to version 16.

### DIFF
--- a/linux/step08_rocky9_config.sh
+++ b/linux/step08_rocky9_config.sh
@@ -3,7 +3,7 @@
 set -e -u -x
 
 #start-recommended
-cp omero-server-systemd.service /etc/systemd/system/omero-server.service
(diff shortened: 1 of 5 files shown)


dominikl:
Forgot that I already had a PR open for that :-)

linux/step08_rocky9_config.sh
```diff
@@ -3,7 +3,7 @@
 set -e -u -x
 
 #start-recommended
-cp omero-server-systemd.service /etc/systemd/system/omero-server.service
+wget https://omero.readthedocs.io/en/stable/_downloads/08850daa9012dfe7b7d4a90dc76c1b83/omero-server-systemd.service -O /etc/systemd/system/omero-server.service
```
jburel: Not sure about that

dominikl: We provide the file so why not use it directly, instead of telling people to download and copy it "manually". But I can remove that change.

jburel: I am thinking more the sha1 instead of something more "generic"

pwalczysko (approved):

